PR #2, opened 2015-01-08
[FIXED JENKINS-23468] Permit multiple variables per binding

jglick:
[JENKINS-23468](https://issues.jenkins-ci.org/browse/JENKINS-23468)


jglick:
@stephenc as the only other person to have touched this plugin, would you like to review?

src/main/java/org/jenkinsci/plugins/credentialsbinding/Binding.java
```diff
@@ -86,25 +70,40 @@ public String getCredentialsId() {
 
     }
 
+    @Deprecated
```
jglick: Signature change in preparation for [JENKINS-26051](https://issues.jenkins-ci.org/browse/JENKINS-26051), while I am refactoring the API anyway.

src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/Messages.properties
```diff
@@ -1,5 +1,6 @@
 FileBinding.secret_file=Secret file
 SecretBuildWrapper.use_secret_text_s_or_file_s_=Use secret text(s) or file(s)
 StringBinding.secret_text=Secret text
-UsernamePasswordBinding.username_and_password=Username and password
+UsernamePasswordBinding.username_and_password=Username and password (conjoined)
+UsernamePasswordMultiBinding.username_and_password=Username and password (separated)
```
jglick: Would prefer to deprecate `UsernamePasswordBinding` but it cannot simply be replaced with a `UsernamePasswordMultiBinding`, and AFAIK we cannot remove the option to add a _new_ `UsernamePasswordBinding` without also breaking reconfiguration of _existing_ jobs, so for now just leaving both options.

pom.xml
```diff
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.532.2</version>
+    <version>1.580.1</version>
```
stephenc: Is this the minimum required version to jump to?

src/main/java/org/jenkinsci/plugins/credentialsbinding/Binding.java
```diff
@@ -24,57 +24,41 @@
 
 package org.jenkinsci.plugins.credentialsbinding;
 
-import com.cloudbees.plugins.credentials.CredentialsProvider;
```
stephenc: Rather noisy diff

stephenc:
Otherwise the BMOTGH says ok

pom.xml
jglick: Yes, at least for upcoming Workflow integration.

src/main/java/org/jenkinsci/plugins/credentialsbinding/Binding.java
jglick: True, but this is an introduce-superclass refactoring, so hard to avoid.

pom.xml
stephenc: Ah so this version bump is not required for this fix, only for a future change

jglick: Yeah, I just did not want to deal with such an old baseline when I knew it was going to be bumped up anyway.

stephenc: My preference is to keep the baseline until forced to update as that way *if* anyone wants to cherry pick backports it is easier

PR #3, opened 2015-01-12
[FIXED JENKINS-26051] withCredentials step

jglick:
[JENKINS-26051](https://issues.jenkins-ci.org/browse/JENKINS-26051)

- [X] basic impl
- [X] simplify SPI
- [X] add automated tests
- [x] run manual testing


albers:
@jglick sorry, I accidentally clicked a checkbox in the issue description, not sure about original values, please verify.

jglick:
@albers no problem, I rechecked it.

@reviewbybees

src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
```diff
@@ -60,21 +63,47 @@ public final String getCredentialsId() {
...
+        }
+
+    }
+
+    /** Callback run at the end of a build. */
+    public interface Unbinder extends Serializable {
         /** Performs any needed cleanup. */
-        void unbind() throws IOException, InterruptedException;
+        void unbind(@Nonnull Run<?,?> build, FilePath workspace, Launcher launcher, TaskListener listener) throws IOException, InterruptedException;
+    }
```
recampbell: I suppose this is an internal interface of which there are no external implementations?

rsandell:
+1

src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
jglick: While I know of no external implementations, this is intended as a _potential_ external interface. At any rate, `MultiBinding` itself was introduced during unreleased changes in #2, so changing it before release cannot be breaking anyone.

PR #5, opened 2015-03-26
[JENKINS-27631] Stronger tests

jglick:
[JENKINS-27631](https://issues.jenkins-ci.org/browse/JENKINS-27631)

I lay awake in bed wondering whether the contextual `EnvVars` got stored somewhere unsafe. Turns out it does, though only temporarily. This PR just makes the test coverage more interesting.

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
@@ -78,11 +78,11 @@ public class BindingStepTest {
     }
 
     @Test public void basics() throws Exception {
+        final String credentialsId = "creds";
+        final String username = "bob";
+        final String password = "s3cr3t";
         story.addStep(new Statement() {
             @Override public void evaluate() throws Throwable {
-                String credentialsId = "creds";
-                String username = "bob";
-                String password = "s3cr3t";
                 UsernamePasswordCredentialsImpl c = new UsernamePasswordCredentialsImpl(CredentialsScope.GLOBAL, credentialsId, "sample", username, password);
                 CredentialsProvider.lookupStores(story.j.jenkins).iterator().next().addCredentials(Domain.global(), c);
                 WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
```diff
@@ -148,24 +171,46 @@
...
+        for (File kid : kids) {
+            String qualifiedName = prefix + kid.getName();
+            if (kid.isDirectory()) {
+                grep(kid, text, qualifiedName + "/", matches);
+            } else if (kid.isFile() && FileUtils.readFileToString(kid).contains(text)) {
+                matches.add(qualifiedName);
+            }
+        }
+    }
+
```
tfennelly: Seems like the above two `grep` methods could be useful utils to make available from the core test-harness.

tfennelly:
:+1:

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
jglick: Perhaps. I cannot recall needing this kind of thing before so I just wrote it as a one-off.

(BTW just the first overload would be `public`—the second is a helper method for the recursion. No easy way in Java to declare a local method, alas.)

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
```diff
@@ -77,24 +78,45 @@ public class BindingStepTest {
...
+                SemaphoreStep.waitForStart("basics/1", b);
+            }
+        });
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = story.j.jenkins.getItemByFullName("p", WorkflowJob.class);
+                assertNotNull(p);
+                WorkflowRun b = p.getBuildByNumber(1);
+                assertNotNull(b);
+                assertEquals("TODO JENKINS-27631", Collections.singleton("program.dat"), grep(b.getRootDir(), password));
```
jglick, line 109: ⇒ #6

PR #8, opened 2015-09-10
Jenkins 30326

jtnord:
added a test case for [JENKINS-30326](https://issues.jenkins-ci.org/browse/JENKINS-30326) and up the dependency to pull in the fix.

@reviewbybees -esp @jglick

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
@@ -29,30 +29,47 @@ import com.cloudbees.plugins.credentials.CredentialsScope;
 import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
 import com.cloudbees.plugins.credentials.domains.Domain;
 import com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl;
+
+import jenkins.security.QueueItemAuthenticatorConfiguration;
+
 import hudson.FilePath;
 import hudson.model.FileParameterValue;
 import hudson.model.Node;
 import hudson.model.Result;
+import hudson.security.FullControlOnceLoggedInAuthorizationStrategy;
 import hudson.slaves.DumbSlave;
 import hudson.slaves.NodeProperty;
 import hudson.slaves.RetentionStrategy;
 import hudson.util.Secret;
+
 import java.io.File;
 import java.io.IOException;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+
 import javax.inject.Inject;
+
 import org.apache.commons.io.FileUtils;
+import org.jenkinsci.plugins.authorizeproject.AuthorizeProjectProperty;
+import org.jenkinsci.plugins.authorizeproject.ProjectQueueItemAuthenticator;
+import org.jenkinsci.plugins.authorizeproject.strategy.SpecificUsersAuthorizationStrategy;
+import org.jenkinsci.plugins.authorizeproject.AuthorizeProjectStrategy;
 import org.jenkinsci.plugins.credentialsbinding.MultiBinding;
 import org.jenkinsci.plugins.plaincredentials.impl.FileCredentialsImpl;
 import org.jenkinsci.plugins.plaincredentials.impl.StringCredentialsImpl;
+import org.jenkinsci.plugins.scriptsecurity.sandbox.Whitelist;
+import org.jenkinsci.plugins.scriptsecurity.sandbox.whitelists.BlanketWhitelist;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.jenkinsci.plugins.workflow.steps.StepConfigTester;
 import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
+
 import static org.junit.Assert.*;
+
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
@@ -60,6 +77,7 @@ import org.junit.runners.model.Statement;
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
 import org.jvnet.hudson.test.RestartableJenkinsRule;
+import org.jvnet.hudson.test.recipes.WithPlugin;
 
 public class BindingStepTest {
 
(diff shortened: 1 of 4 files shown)


amuniz:
:bee: (I'd love to have a link to the issue in the description)

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
```diff
@@ -240,6 +258,49 @@
...
+                        + "  }\n"
+                        + "}", true));
+                // run the job as a specific user
+                p.addProperty(new AuthorizeProjectProperty(new SpecificUsersAuthorizationStrategy("dummy", true)));
+
+                // the build will fail if we can not locate the credentials
+                WorkflowRun b = story.j.assertBuildStatusSuccess(p.scheduleBuild2(0).get());
+                // make sure this was actually run as a user and not system
+                story.j.assertLogContains("running as user: dummy", b);
+            }
```
tfennelly: Should it be testing for the `echo`d text from inside the `withCredentials` closure?

jtnord: see the contents of those steps and the assert on line 297

tfennelly: Fair enough but, while you're at it, I'd still add a simple log contains test to make sure it doesn't fail silently. That should already be covered by the tests for `withCredentials`, but it wouldn't hurt adding it here too (one simple line).

tfennelly:
:bee:

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
```diff
@@ -240,6 +258,53 @@
...
+                // configure security
+                story.j.jenkins.setSecurityRealm(story.j.createDummySecurityRealm());
+                story.j.jenkins.setAuthorizationStrategy(new FullControlOnceLoggedInAuthorizationStrategy());
+                // create the user.
+                story.j.jenkins.getUser("dummy");
+                
+                // enable the run as user strategy for the AuthorizeProject plugin
+                Map<String, Boolean> strategies = new HashMap<String, Boolean>();
+                strategies.put(story.j.jenkins.getDescriptor(SpecificUsersAuthorizationStrategy.class).getId(), true);
+                QueueItemAuthenticatorConfiguration.get().getAuthenticators().add(new ProjectQueueItemAuthenticator(strategies));
```
oleg-nenashev: Makes sense to explicitly enable the strategy in AuthorizeProject global configs. It's disabled by default => the behavior may change at some point when the plugin has more checks

oleg-nenashev:
:bee: with a minor comment

rsandell:
:bee:

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
jtnord: any change would be caught by the assertion on line 303  `assertLogContains("running as user: dummy", b);`
FWIW this is how the tests in AuthorizeProject are written.

jtnord:
@reviewbybees done (oops)

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
jglick: BTW it is common to directly implement a particular `QueueItemAuthenticator` as a test extension, so that you do not need to actually depend on `authorize-project`. But this is fine too, of course—just makes it a bit more like an integration test and a bit less like a unit test.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
```diff
@@ -240,6 +258,53 @@
...
+                CredentialsProvider.lookupStores(story.j.jenkins).iterator().next().addCredentials(Domain.global(), new StringCredentialsImpl(CredentialsScope.GLOBAL, credentialsId, "sample", Secret.fromString(secret)));
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+
+                p.setDefinition(new CpsFlowDefinition(""
+                        + "node {\n"
+                        + "  def authentication = Jenkins.getAuthentication()\n"
+                        + "  echo \"running as user: $authentication.principal\"\n"
+                        + "  withCredentials([[$class: 'StringBinding', credentialsId: '" + credentialsId + "', variable: 'SECRET']]) {\n"
+                        + "    writeFile file:'test', text: \"$env.SECRET\"\n"
+                        + "    def content = readFile 'test'\n"
```
jglick:
```groovy
def context = env.SECRET
```

seems simpler, no?

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/UsernamePasswordBindingTest.java
```diff
@@ -52,7 +58,7 @@
         FreeStyleProject p = r.createFreeStyleProject();
         p.getBuildWrappersList().add(new SecretBuildWrapper(Collections.<Binding<?>>singletonList(new UsernamePasswordBinding("AUTH", c.getId()))));
         p.getBuildersList().add(new Shell("set +x\necho $AUTH > auth.txt"));
-        r.configRoundtrip(p);
+        r.configRoundtrip((Item)p);
```
jglick: Why?

jglick: Ah, d3abea3b4c8b66dd42d022ff14ecb0482f044291, sigh.

jtnord: ambiguous method call.

`FreeStyleProject` is an `Item` as well as a `Project` so the call is [ambiguous](https://github.com/jenkinsci/jenkins/blob/7adbeef5a70a37768f6d2f8caf3f9e29c8e57436/test/src/main/java/org/jvnet/hudson/test/JenkinsRule.java#L1041-L1049)

the code will not compile in eclipse without this - I thought I removed this call from Jenkins Rule a while ago and was just using an older version - but perhaps not.

jtnord: see jenkinsci/jenkins#1830

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
jtnord: now you point it out that appears completely obvious :-)

PR #9, opened 2015-10-14
Fix resource leak identified by eclipse

jtnord:
@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBinding.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBinding.java
@@ -26,13 +26,17 @@ package org.jenkinsci.plugins.credentialsbinding.impl;
 
 import com.cloudbees.plugins.credentials.CredentialsProvider;
 import com.cloudbees.plugins.credentials.domains.DomainRequirement;
+
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.model.Item;
 import hudson.util.FormValidation;
+
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Collections;
+
+import org.apache.commons.io.IOUtils;
 import org.jenkinsci.plugins.credentialsbinding.BindingDescriptor;
 import org.jenkinsci.plugins.plaincredentials.FileCredentials;
 import org.kohsuke.stapler.AncestorInPath;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBinding.java
```diff
@@ -73,6 +78,11 @@ public FormValidation doCheckCredentialsId(@AncestorInPath Item owner, @QueryPar
                     } catch (IOException x) {
                         return FormValidation.warning("Could not verify file format");
                     }
+                    finally {
+                        if (is != null) {
+                            IOUtils.closeQuietly(is);
```
jglick: :ant: better style is to use

```java
try {
    InputStream is = c.getContent();
    try {
        // use it…
    } finally {
        is.close();
    }
} catch (IOException x) {
    return FormValidation.warning("Could not verify file format");
}
```

since this avoids use of a non-blank-final variable as well as the undesirable `closeQuietly`, and allows for easy mechanical translation (e.g., IDE hint) to the right idiom when you upgrade to `-source 7` or newer:

```java
try {
    try (InputStream is = c.getContent()) {
        // use it…
    }
} catch (IOException x) {
    return FormValidation.warning("Could not verify file format");
}
```

jglick:
:bee:

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBinding.java
```diff
@@ -73,6 +78,11 @@ public FormValidation doCheckCredentialsId(@AncestorInPath Item owner, @QueryPar
                     } catch (IOException x) {
                         return FormValidation.warning("Could not verify file format");
                     }
+                    finally {
+                        if (is != null) {
```
stephenc: `IOUtils.closeQuietly(x)` does a null check so this if clause is unnecessary

jtnord:
@reviewbybees done

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBinding.java
KostyaSha:
```
try (InputStream is = c.getContent()) {
        // use it…
} catch (IOException x) {
    return FormValidation.warning("Could not verify file format");
}
```

PR #1, opened 2014-08-27
[FIXED JENKINS-29255] Restricting secret file visibility to avoid "WARNING: UNPROTECTED PRIVATE KEY FILE!" when using as ssh key

Lohandus:
[JENKINS-29255](https://issues.jenkins-ci.org/browse/JENKINS-29255)

PR #10, opened 2015-10-15
[FIXED JENKINS-30941] do not set directories to 0400 on unix if you want to delete their contents...

jtnord:
`ZipFileBinding` was not working correctly as it caused errors when the step was exited as the directory could not be cleans up on Unix like OSes.

caused by #1 

[JENKINS-30941](https://issues.jenkins-ci.org/browse/JENKINS-30941)

@reviewbybees


pom.xml
```diff
@@ -18,7 +18,7 @@
   </description>
   <url>http://wiki.jenkins-ci.org/display/JENKINS/Credentials+Binding+Plugin</url>
   <properties>
-    <workflow.version>1.5</workflow.version>
+    <workflow.version>1.7</workflow.version>
```
jtnord: without this the unit test times out as the error is not caught by workflow and it stays in a zombie building state.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBindingTest.java
```diff
@@ -0,0 +1,80 @@
...
+        WorkflowJob p = j.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition(""
+                                              + "node {\n"
+                                              + "  withCredentials([[$class: 'ZipFileBinding', credentialsId: 'zipfile', variable: 'ziploc']]) {\n"
+                                              + (Functions.isWindows() 
+                                                      ? "    bat 'type %ziploc%\\\\META-INF\\\\MANIFEST.MF'\n"
+                                                      : "    sh 'cat ${ziploc}/META-INF/MANIFEST.MF'\n" )
+                                              + "    def manifest = readFile encoding: 'UTF-8', file: \"${env.ziploc}/META-INF/MANIFEST.MF\"\n"
+                                              + "    echo \"manifest contents....\\n${manifest}\"\n"
+                                              + "    if (!manifest.contains('Specification-Title: Apache Commons FileUpload')) {\n"
```
amuniz: Test depending on a transitive dependency. If that dependency is upgraded and the `Specification-Title` is changed, the test will fail. Why not choose a fixed dependency set in this plugin pom? It would be safer.

However, it's improbable that `Specification-Title` changes, so just noting it.

amuniz:
:bee:

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBindingTest.java
jtnord: Yes and I would tend to agree with this.  However the non transative dependencies are Jenkins plugins and their manifest is a little sparse and not what I would call stable (the implementation-version is just the groupId).

But now I mention that I could have used the META-INF/maven/${groupId}/${artifactId}/pom.properties for workflow-api which should have a stable group and version...

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBindingTest.java
```diff
@@ -0,0 +1,80 @@
...
+import hudson.Functions;
+import hudson.model.FileParameterValue.FileItemImpl;
+
+public class ZipFileBindingTest {
+
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+
+    @Issue("30941")
```
jglick: :ant: should be

```java
@Issue("JENKINS-30941")
```

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBindingTest.java
jglick: :ant: this is all unnecessary. Just make a tiny test ZIP file and add to `src/test/resources/`.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBindingTest.java
```diff
@@ -0,0 +1,80 @@
...
+    @Test
+    public void cleanUpSucceeds() throws Exception {
+        /** Issue was just present on Linux not windows - but the test will run on both */
+
+        final String credentialsId = "zipfile";
+        // grab a jar file here - just use commons.fileupload!
+        File jar = new File(FileItem.class.getProtectionDomain().getCodeSource().getLocation().toURI());
```
jglick: :ant: should not assume that classpath entries happen to be inside file directories (rather than themselves inside JARs). Always use `Class[Loader].getResourceAsStream` and copy to a temp file (`@Rule TemporaryFolder`).

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBindingTest.java
```diff
@@ -0,0 +1,80 @@
...
+                                                      ? "    bat 'type %ziploc%\\\\META-INF\\\\MANIFEST.MF'\n"
+                                                      : "    sh 'cat ${ziploc}/META-INF/MANIFEST.MF'\n" )
+                                              + "    def manifest = readFile encoding: 'UTF-8', file: \"${env.ziploc}/META-INF/MANIFEST.MF\"\n"
+                                              + "    echo \"manifest contents....\\n${manifest}\"\n"
+                                              + "    if (!manifest.contains('Specification-Title: Apache Commons FileUpload')) {\n"
+                                              + "      error ('incorrect details from zip file')\n"
+                                              + "    }\n"
+                                              + "  }\n"
+                                              + "}\n"
+                                              , false /* String.contains is not whitelisted until script-security 1.15. */));
```
jglick: You can set a `test`-scoped dependency, or more simply just have the flow print out the complete manifest and then use `JenkinsRule.assertBuildLogContains`.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBindingTest.java
```diff
@@ -0,0 +1,80 @@
...
+        FileCredentialsImpl fc = new FileCredentialsImpl(CredentialsScope.GLOBAL, credentialsId, "Just a zip file", fi, fi.getName(), null);
+
+        CredentialsProvider.lookupStores(j.jenkins).iterator().next().addCredentials(Domain.global(), fc);
+        WorkflowJob p = j.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition(""
+                                              + "node {\n"
+                                              + "  withCredentials([[$class: 'ZipFileBinding', credentialsId: 'zipfile', variable: 'ziploc']]) {\n"
+                                              + (Functions.isWindows() 
+                                                      ? "    bat 'type %ziploc%\\\\META-INF\\\\MANIFEST.MF'\n"
+                                                      : "    sh 'cat ${ziploc}/META-INF/MANIFEST.MF'\n" )
```
jglick: What is the purpose of this line in the flow script?

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBindingTest.java
jtnord: this is a JAR - it gets the path to the jar file.

jglick: Only if the test harness you are using happens to load resources from `target/test-classes/` plus other filesystem locations. Obviously `mvn surefire:test` will, but IDEs are not obliged to. Just use the safe idiom, it is easy anyway.

jglick:
:bee: despite the unnecessarily fragile test.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBindingTest.java
jtnord: its testing that the `ZipFileBinding` actually works - the variable is set correctly to the secret directory and you can read the unpacked file (which was a different issue and not reported - but `ZipFileBinding` was DOA in 1.5 on Linux as you could not read the secret files) - yes could/should be a separate test but I still makes sense to test the whole `ZipFileBinding` can bind/unbind inject the location into the context and that the unpacked files can actually be read!

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBindingTest.java
jtnord: I don't follow.  Yes if your IDE has commons-FileUpload open and links projects classpaths it will get a directory rather than a jar - but then `getResourceAsStream()` will get a class file and not a jar.

jtnord: this class (FileItem) is not in this project - it is from a different maven project so will never be in target/test-classes.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBindingTest.java
jglick: Right, but the `readFile` step in the next line is already testing that. I am asking what this `cat`/`type` line is adding.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBindingTest.java
jglick: > then `getResourceAsStream()` will get a class file and not a jar

Which is why you need to create a test ZIP file which you can add to `src/test/resources/` to be sure that `getResourceAsStream()` will return a ZIP byte sequence you can copy to a local file.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBindingTest.java
jtnord: not a huge amount but it takes milliseconds to execute.
I was also hitting a very strange workflow bug - which you have fixed in 1.11-SNAPSHOT so hence I went for both.

oleg-nenashev:
:bee:

jtnord:
@reviewbybees done

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBindingTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Issue("JENKINS-30941")
+    @Test
+    public void cleanUpSucceeds() throws Exception {
+        /** Issue was just present on Linux not windows - but the test will run on both */
+
+        final String credentialsId = "zipfile";
+
+        /* do the dance to get a simple zip file into jenkins */
+        InputStream zipStream = this.getClass().getResourceAsStream("a.zip");
```
jglick: BTW FindBugs may warn you that this is an antipattern which would break if you ever subclassed this class from a different package (obviously not likely in this case); proper style is to use `ZipFileBindingTest.class.getResourceAsStream("a.zip")`.

PR #11, opened 2016-02-15
Certificate binding to variables

idstein:
- Alias of certificate name (assumed to be description of certificate)
- Password to access certificate keystore
- Keystore file location (temporary keystore deleted after each run)


idstein:
Whoops looks like my bash foo has seen, some better times. I will check back on the failing test later, when I have access to my linux machine.

On Windows it looked okay.

Maybe, I will add also a check on that the certificate is going to be deleted. Just to be sure.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding.java
```diff
@@ -0,0 +1,131 @@
...
+		} else {
+			return new MultiEnvironment(m);
+		}
+	}
+
+	private static class UnbinderImpl implements Unbinder {
+
+		private static final long serialVersionUID = 1;
+
+		private final FilePath keyStoreFile;
```
jglick: This is not safe to serialize. See `FileBinding` for how to do it.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding.java
```diff
@@ -0,0 +1,131 @@
...
+	public String getAliasVariable() {
+		return aliasVariable;
+	}
+
+	private final String passwordVariable;
+	private final String aliasVariable;
+
+	@DataBoundConstructor
+	public CertificateMultiBinding(String keystoreVariable, String passwordVariable, String aliasVariable,
+			String credentialsId) {
```
jglick: `keystoreVariable` is likely mandatory, but `passwordVariable` and `aliasVariable` seem optional, so it would be better to specify these with `@DataBoundSetter`. Would allow them to be omitted in a Pipeline job using `withCredentials`.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/FileBinding.java
```diff
@@ -85,7 +85,7 @@
         
     }
 
-    private static FilePath secretsDir(FilePath workspace) {
+    static FilePath secretsDir(FilePath workspace) {
```
jglick: Needs merge with `master`.

src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding/config-variables.jelly
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+The MIT License
+
+Copyright 2013 jglick.
```
jglick: You can use your own name. :-)

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding.java
idstein: I was also not sure wether this is safe for serialisation or not, but FilePath was stated by the Jenkins documentation to implement Serializable, so I assumed that would be safe.

So I simply have to call getRemote to transform it?

src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding/config-variables.jelly
idstein: Agree that was a straight copy & paste, my bad!

idstein:
Any update on the review?

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding.java
jglick: > FilePath was stated by the Jenkins documentation to implement Serializable

Only over a remoting channel, which is not the relevant case here.

jglick: > call getRemote

Yes that should work. In `unbind` you need to reconstruct the live `FilePath` using `workspace` since it may be running on an agent.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding.java
```diff
@@ -0,0 +1,142 @@
...
+		final String storePassword = credentials.getPassword().getPlainText();
+		Map<String, String> m = new HashMap<String, String>();
+		if(aliasVariable!=null && !aliasVariable.isEmpty())
+			m.put(aliasVariable, credentials.getDescription());
+		if(passwordVariable!=null && !passwordVariable.isEmpty())
+			m.put(passwordVariable, storePassword);
+
+		if (workspace != null) {
+			FilePath secrets = FileBinding.secretsDir(workspace);
+			final String tempKeyStoreName = UUID.randomUUID().toString();
```
jglick: Should be rewritten to use #25.

PR #12, opened 2016-02-15
[JENKINS-24805] Mask all bindings with asterix

idstein:
- Order bindings by largest to shortest string to cover substrings
- Fix an issue if secret is partially duplicated, e.g. "android"
  "androiddebugkey" which results in "***_" and "**_*debugkey"

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
@@ -125,7 +125,7 @@ public final class BindingStep extends AbstractStepImpl {
     }
 
     /** Similar to {@code MaskPasswordsOutputStream}. */
-    private static final class Filter extends ConsoleLogFilter implements Serializable {
+    static final class Filter extends ConsoleLogFilter implements Serializable {
 
         private static final long serialVersionUID = 1;
 
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
@@ -40,6 +40,8 @@ import java.io.OutputStream;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -130,10 +132,19 @@ public final class BindingStep extends AbstractStepImpl {
         private static final long serialVersionUID = 1;
 
         private final Secret pattern;
+        
+        private static final Comparator<String> StringLengthComparator = new Comparator<String>() {
+			@Override
+			public int compare(String o1, String o2) {
+				return o1.length() - o2.length();
+			}
+		};
 
         Filter(Collection<String> secrets) {
             StringBuilder b = new StringBuilder();
-            for (String secret : secrets) {
+            List<String> hiddenSecrets = new ArrayList<String>(secrets);
+            hiddenSecrets.sort(StringLengthComparator.reversed());
+            for (String secret : hiddenSecrets) {
                 if (b.length() > 0) {
                     b.append('|');
                 }
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
@@ -143,7 +143,8 @@ public final class BindingStep extends AbstractStepImpl {
         Filter(Collection<String> secrets) {
             StringBuilder b = new StringBuilder();
             List<String> hiddenSecrets = new ArrayList<String>(secrets);
-            hiddenSecrets.sort(StringLengthComparator.reversed());
+            Collections.sort(hiddenSecrets, StringLengthComparator);
+            Collections.reverse(hiddenSecrets);
             for (String secret : hiddenSecrets) {
                 if (b.length() > 0) {
                     b.append('|');
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/FileBinding.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/FileBinding.java
@@ -50,7 +50,9 @@ public class FileBinding extends Binding<FileCredentials> {
     }
 
     @Override public SingleEnvironment bindSingle(Run<?,?> build, FilePath workspace, Launcher launcher, TaskListener listener) throws IOException, InterruptedException {
-        FileCredentials credentials = getCredentials(build);
+        if(workspace==null)
+        	return new SingleEnvironment("");
+    	FileCredentials credentials = getCredentials(build);
         FilePath secrets = secretsDir(workspace);
         String dirName = UUID.randomUUID().toString();
         final FilePath dir = secrets.child(dirName);
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
@@ -95,10 +95,8 @@ public class SecretBuildWrapper extends BuildWrapper {
     		return super.decorateLogger(build, logger);
     	
     	Map<String,String> overrides = new HashMap<String,String>();
-        List<MultiBinding.Unbinder> unbinders = new ArrayList<MultiBinding.Unbinder>();
         for (MultiBinding<?> binding : bindings) {
-            MultiBinding.MultiEnvironment environment = binding.bind(build, build.getWorkspace(), null, null);
-            unbinders.add(environment.getUnbinder());
+            MultiBinding.MultiEnvironment environment = binding.bind(build, null, null, null);
             overrides.putAll(environment.getValues());
         }
         return new BindingStep.Filter(overrides.values()).decorateLogger(build, logger);
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
@@ -29,10 +29,13 @@ import hudson.Launcher;
 import hudson.model.AbstractBuild;
 import hudson.model.AbstractProject;
 import hudson.model.BuildListener;
+import hudson.model.Run.RunnerAbortedException;
 import hudson.tasks.BuildWrapper;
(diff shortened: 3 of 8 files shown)


idstein:
Some tests showed that this is a breaking change to Stapler framework. If we want to have that checkbox, it must be be capable of migrating existing jobs to this this new SecretBuildWrapprr using the newInstance method on its Descriptor.

jglick:
I suppose this is trying to solve [JENKINS-24805](https://issues.jenkins-ci.org/browse/JENKINS-24805)?

src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper/config.jelly
```diff
@@ -29,4 +29,9 @@ THE SOFTWARE.
             <f:repeatableHeteroProperty field="bindings" hasHeader="true"/>
         </f:block>
     </f:section>
+    <f:advanced>
+    	<f:entry title="Show bindings" field="showBindings" description="If checked, all secret variable bindings will be clear-text.">
```
jglick: Why is this even an option? I would just delete it. `BindingStep` does not offer such an option, and AFAIK no one has ever asked for it.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -71,6 +87,22 @@
...
     }
+    
+    @Override
+    public OutputStream decorateLogger(AbstractBuild build, OutputStream logger)
+    		throws IOException, InterruptedException, RunnerAbortedException {
+    	if(showBindings)
+    		return super.decorateLogger(build, logger);
+    	
+    	Map<String,String> overrides = new HashMap<String,String>();
+        List<MultiBinding.Unbinder> unbinders = new ArrayList<MultiBinding.Unbinder>();
```
jglick: Why are you collecting these? The result is unused.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -71,6 +87,22 @@
...
+        for (MultiBinding<?> binding : bindings) {
+            MultiBinding.MultiEnvironment environment = binding.bind(build, build.getWorkspace(), null, null);
```
jglick: Do not call `bind` as it will actually do things, like copy files, which `setUp` already did. Just call `binding.variables()`, as `makeSensitiveBuildVariables` does.

src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper/config.jelly
idstein: I though that it would be really useful for debugging especially, if you want to see if there are any errors related to your password having a typo or similar or a path that is not accessible at all.

The old behaviour would is to show all bindings, so that option would show the same behaviour as before.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
idstein: As referred in the next review comment, I thought it would be necessary to call the unbind step after calling the bind step, but that is not necessary anymore after changing the bind step to not do any actual binding, when called with an invalid launcher and workspaces (is null at this stage, but may change in future).

idstein:
In the future Jenkins core must not only mask sensitive variables for Ant and Maven but also provide a proper interface for all sort of runners such as Gradle, Shell or Batch.

For now, using the bindings with an invalid workspaces may produce a proper variable value set. This in turn can be used to decorate the logger and avoid the lacking of passwords. However, this may also mask false positive in the logs such as an arbitrary log statement using the secrete variable value.

src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper/config.jelly
daniel-beck: I agree with Jesse. This looks a lot like YAGNI. If there's actual demand (and not just one person), this can always be added later.

idstein: If you do not need it, you don't have to check it. It's unit tested and does not do any harm due to it's simple logic.

I do not see how this contradicts the Jenkins contribution guidance.

idstein:
Any update on this open pull request?

src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper/config.jelly
jglick: > does not do any harm

Except that once added, it cannot be removed.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -125,15 +127,25 @@
...
+				return o1.length() - o2.length();
+			}
+		};
 
         Filter(Collection<String> secrets) {
             StringBuilder b = new StringBuilder();
-            for (String secret : secrets) {
+            List<String> hiddenSecrets = new ArrayList<String>(secrets);
+            Collections.sort(hiddenSecrets, StringLengthComparator);
+            Collections.reverse(hiddenSecrets);
```
jglick: Why would you not just invert the comparator?

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -125,15 +127,25 @@
...
+			public int compare(String o1, String o2) {
+				return o1.length() - o2.length();
+			}
+		};
 
         Filter(Collection<String> secrets) {
             StringBuilder b = new StringBuilder();
-            for (String secret : secrets) {
+            List<String> hiddenSecrets = new ArrayList<String>(secrets);
+            Collections.sort(hiddenSecrets, StringLengthComparator);
```
jglick: Is this intended as some sort of performance optimization for the regexp, or what?

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/FileBinding.java
```diff
@@ -49,7 +49,9 @@
     }
 
     @Override public SingleEnvironment bindSingle(Run<?,?> build, FilePath workspace, Launcher launcher, TaskListener listener) throws IOException, InterruptedException {
-        FileCredentials credentials = getCredentials(build);
+        if(workspace==null)
```
jglick: Huh? `workspace` should be considered `@Nonnull`. Anyway what does this have to do with the rest of the PR?

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -29,28 +29,44 @@
...
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 
 @SuppressWarnings({"rawtypes", "unchecked"}) // inherited from BuildWrapper
 public class SecretBuildWrapper extends BuildWrapper {
 
     private final List<? extends MultiBinding<?>> bindings;
+    private boolean showBindings;
 
+    public SecretBuildWrapper(List<? extends MultiBinding<?>> bindings, boolean showBindings) {
```
jglick: Unnecessary, delete.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -29,28 +29,44 @@
...
+        this.bindings = bindings;
+        this.showBindings = showBindings;
+    }
+    
     @DataBoundConstructor public SecretBuildWrapper(List<? extends MultiBinding<?>> bindings) {
         this.bindings = bindings;
+        showBindings = true;
```
jglick: Why? The setter will be called immediately afterwards anyway.

And if you meant to retain former behavior for projects loaded from disk for some reason (I do not recommend it), this does not do it. You would need to add a `readResolve`.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -71,6 +87,28 @@
...
+    
+    @Override
+    public OutputStream decorateLogger(AbstractBuild build, OutputStream logger)
+    		throws IOException, InterruptedException, RunnerAbortedException {
+    	if(showBindings)
+    		return super.decorateLogger(build, logger);
+    	
+    	Map<String,String> overrides = new HashMap<String,String>();
+        for (MultiBinding<?> binding : bindings) {
+            MultiBinding.MultiEnvironment environment = binding.bind(build, null, null, null);
```
jglick: No, do not do this. `bind` should be called only once, during `setUp`.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
jglick: Or something to do with substrings? Should be done as a separate PR. Nothing to do with the rest of this PR.

src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper/config.jelly
jglick: > Especially, as sensitive variables are only masked for Ant or Maven based builds. That means any command line like build step such as Gradle may fail with a hidden whitespace exception via its parameters.

What does this mean?

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/FileBinding.java
jglick: I guess it is a side effect of calling `bind` with null values, which you should not do.

escoem:
Hello @idstein , I'm interested in finishing the PR so if you aren't going to do and you agree I will address all stuff commented by @jglick and @daniel-beck

escoem:
Hi, I'm ready for the new PR.

idstein:
My bad, I didn't find time to work on this. Of course, you can go ahead and finish this issue.

escoem:
Great @idstein , the new PR will be focused only in the asterisk.

jglick:
@abayer said he would try to do this the right way.

PR #14, opened 2016-02-24
Mask password with *******

harcher81:
And support of CloudBees AWS Credentials Plugin as credential provider.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,6 +55,16 @@
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>aws-credentials</artifactId>
+      <version>1.11</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>envinject</artifactId>
+      <version>1.92.1</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>plain-credentials</artifactId>
       <version>1.0</version>
     </dependency>
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
@@ -24,17 +24,14 @@
 
 package org.jenkinsci.plugins.credentialsbinding;
 
-import com.cloudbees.plugins.credentials.CredentialsDescriptor;
 import com.cloudbees.plugins.credentials.CredentialsProvider;
 import com.cloudbees.plugins.credentials.common.IdCredentials;
 import com.cloudbees.plugins.credentials.common.StandardCredentials;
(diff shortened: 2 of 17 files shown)


pom.xml
```diff
@@ -55,6 +55,16 @@
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>aws-credentials</artifactId>
```
jglick: Really do not want this dependency. If anything, `aws-credentials` could depend on the more generic `credentials-binding`. Anyway this is unrelated to the main thrust of the PR.

pom.xml
```diff
@@ -55,6 +55,16 @@
...
+      <version>1.11</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>envinject</artifactId>
```
jglick: Why?

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -148,7 +148,7 @@
                 @Override protected void eol(byte[] b, int len) throws IOException {
                     Matcher m = p.matcher(new String(b, 0, len));
                     if (m.find()) {
-                        logger.write(m.replaceAll("****").getBytes());
+                        logger.write(m.replaceAll("********").getBytes());
```
jglick: Why?

src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
```diff
@@ -97,17 +94,17 @@ public Unbinder getUnbinder() {
     /** Callback run at the end of a build. */
     public interface Unbinder extends Serializable {
         /** Performs any needed cleanup. */
-        void unbind(@Nonnull Run<?,?> build, FilePath workspace, Launcher launcher, TaskListener listener) throws IOException, InterruptedException;
+        void unbind(@Nonnull Run<?,?> build, FilePath workspace) throws IOException, InterruptedException;
```
jglick: Incompatible.

jglick:
Seems like #12 is already covering this?

harcher81:
Yes, It coverages the same problems, but I think, this is a better solution. In a "Username and password (conjoined)" binding step, if you inject Foo:Bar, with this fix, this hide all occurrences of Bar and not just Foo:bar.

jglick:
> In a "Username and password (conjoined)" binding step, if you inject Foo:Bar, with this fix, this hide all occurrences of Bar and not just Foo:bar

OK…but the plugin was not setting any variable to `Bar` to begin with. It was only setting `Foo:Bar`, which we mask in case that was passed e.g. to `curl -u $USERPASS …` and the shell was in echo mode. Of course some poorly implemented command might be printing the password later for some reason, but if this is a problem you need merely switch to `UsernamePasswordMultiBinding`.

PR #15, opened 2016-03-01
[JENKINS-27387] NPE from Snippetizer

jglick:
[JENKINS-27387](https://issues.jenkins-ci.org/browse/JENKINS-27387)

@reviewbybees esp. @stephenc who is responsible for the weirdness in `select.jelly`.

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep/config.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep/config.jelly
@@ -25,6 +25,7 @@ THE SOFTWARE.
 
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <j:set var="expressionAllowed" value="false"/>
     <f:section title="Bindings">
         <f:block>
             <f:repeatableHeteroProperty field="bindings" hasHeader="true"/>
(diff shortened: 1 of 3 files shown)


src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep/config.jelly
```diff
@@ -25,6 +25,7 @@ THE SOFTWARE.
 
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <j:set var="expressionAllowed" value="false"/>
```
stephenc: why is an expression not allowed here? or is it this that causes the NPE and you are just hacking around the NPE

jglick: Both, really.

First, passing `true` causes an NPE, which ultimately is due to the fact that the POSTed JSON looks like

```json
"credentialsId":["mycreds",""]
```

rather than the expected

```json
"credentialsId":"mycreds"
```

which indicates that `buildFormTree` is choking on the way `select.jelly` is written. Frankly I am not sure how it works for the freestyle `SecretBuildWrapper`; somehow the correct JSON is sent in that case.

Second, for `BindingStep`, the configuration form is only displayed in _Snippet Generator_ for a Pipeline job. In this case there is no advantage to allowing an expression, and indeed it is just misleading and confusing. _Any_ parameter to a step which you want to make dynamic in some way you can do so using normal Groovy code; there is no reason for this one to be treated differently.

stephenc:
:bee:

src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep/config.jelly
jglick: > somehow the correct JSON is sent in that case

…or not: [JENKINS-37707](https://issues.jenkins-ci.org/browse/JENKINS-37707)

PR #16, opened 2016-05-24
[JENKINS-35095] Findbugs issues. Migrate to 2.9 parent pom.

armfergom:
- Migrate to 2.9 parent pom
- Findbugs issues
- Injected test fix

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,10 +5,9 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.596.1</version>
+    <version>2.9</version>
   </parent>
 
-  <groupId>org.jenkins-ci.plugins</groupId>
   <artifactId>credentials-binding</artifactId>
   <version>1.8-SNAPSHOT</version>
   <packaging>hpi</packaging>
@@ -19,7 +18,9 @@
   <url>http://wiki.jenkins-ci.org/display/JENKINS/Credentials+Binding+Plugin</url>
   <properties>
     <workflow.version>1.7</workflow.version>
-  </properties>
+    <jenkins.version>1.596.1</jenkins.version>
+    <java.level>6</java.level>
+  </properties>  
   <licenses>
     <license>
       <name>MIT</name>
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/Binding.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/Binding.java
@@ -25,12 +25,15 @@
 package org.jenkinsci.plugins.credentialsbinding;
 
 import com.cloudbees.plugins.credentials.common.StandardCredentials;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.FilePath;
 import hudson.Launcher;
 import hudson.Util;
 import hudson.model.AbstractBuild;
 import hudson.model.BuildListener;
 import java.io.IOException;
+
 import javax.annotation.Nonnull;
 
 import hudson.model.Run;
@@ -110,9 +113,13 @@ public abstract class Binding<C extends StandardCredentials> extends MultiBindin
             throw new AbstractMethodError("you must override bindSingle");
         }
     }
+    
     private static class UnbinderWrapper implements Unbinder {
         private static final long serialVersionUID = 1; // only really serialize if what it wraps is, too
+        
+        @SuppressFBWarnings(value="SE_BAD_FIELD", justification="Environment is deprecated and will generally be not serializable")
         private final Environment e;
+        
         UnbinderWrapper(Environment e) {
             this.e = e;
         }
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
@@ -36,6 +36,7 @@ import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.util.Secret;
 import java.io.IOException;
+import java.io.ObjectStreamException;
 import java.io.OutputStream;
 import java.io.Serializable;
 import java.util.ArrayList;
@@ -45,6 +46,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+
+import org.apache.commons.codec.Charsets;
 import org.jenkinsci.plugins.credentialsbinding.MultiBinding;
 import org.jenkinsci.plugins.workflow.steps.AbstractStepDescriptorImpl;
 import org.jenkinsci.plugins.workflow.steps.AbstractStepExecutionImpl;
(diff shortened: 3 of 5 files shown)


src/main/java/org/jenkinsci/plugins/credentialsbinding/Binding.java
```diff
@@ -77,7 +79,7 @@ public SingleEnvironment(String value, Unbinder unbinder) {
     }
 
     @Deprecated
-    public interface Environment {
+    public interface Environment extends Serializable {
```
jglick: Why?

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -146,9 +148,9 @@
             final Pattern p = Pattern.compile(pattern.getPlainText());
             return new LineTransformationOutputStream() {
                 @Override protected void eol(byte[] b, int len) throws IOException {
-                    Matcher m = p.matcher(new String(b, 0, len));
+                    Matcher m = p.matcher(new String(b, 0, len, Charsets.UTF_8));
```
jglick: :bug: this should probably be using `Run.getCharset`. (As noted in https://github.com/jenkinsci/jenkins/pull/2302, it is hard to test these things realistically.)

src/main/java/org/jenkinsci/plugins/credentialsbinding/Binding.java
armfergom: `UnbinderWrapper` has a [field](https://github.com/jenkinsci/credentials-binding-plugin/blob/master/src/main/java/org/jenkinsci/plugins/credentialsbinding/Binding.java#L115) of type `Environment`. `UnbinderWrapper` implements `Unbinder` and `Unbinder` implements `Serializable`. `Environment` interface does not implement `Serializable` and Findbugs complains about that with the following error:

> [INFO] Class org.jenkinsci.plugins.credentialsbinding.Binding$UnbinderWrapper defines non-transient non-serializable instance field e [org.jenkinsci.plugins.credentialsbinding.Binding$UnbinderWrapper] In Binding.java SE_BAD_FIELD

jglick: You just need to suppress that warning. `Environment` is deprecated for a reason: it was not, in general, serializable. `UnbinderWrapper` is used only if a `Binding` subtype failed to be updated to support Pipeline, still implementing `bind`, and such implementations would typically have an unserializable environment.

armfergom:
@jglick addressed your comments. 

- `Environment` no longer extends `Serializable` and Findbugs error is ignored.
- Charset is retrieved from the `Run`. Since the `AbstractBuild`in `decorateLogger` is `null` (as stated [here](https://github.com/jenkinsci/workflow-step-api-plugin/blob/master/src/main/java/org/jenkinsci/plugins/workflow/steps/BodyInvoker.java#L109)), the charset name needs to be passed as parameter in `Filter` constructor. Nor the `Run` or the `Charset` can be a field in `Filter` class since they are not serializable, causing problems when executing the step. Please, advise if you had something different in mind.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -140,15 +143,16 @@
...
             pattern = Secret.fromString(b.toString());
+            this.charsetName = charsetName;
         }
 
         @Override public OutputStream decorateLogger(AbstractBuild _ignore, final OutputStream logger) throws IOException, InterruptedException {
             final Pattern p = Pattern.compile(pattern.getPlainText());
             return new LineTransformationOutputStream() {
                 @Override protected void eol(byte[] b, int len) throws IOException {
-                    Matcher m = p.matcher(new String(b, 0, len));
+                    Matcher m = p.matcher(new String(b, 0, len, Charset.forName(charsetName)));
```
jglick: :bug: this will throw an `IllegalArgumentException("Null charset name")` when resuming a build that had been running while you restarted Jenkins to update this plugin.

In general, when introducing a field to a `Serializable` class, you must consider the possibility that it resolves to `null`.

armfergom: I have a fix for this. But before pushing it I wanted to test it. I have not been able to reproduce the issue at all with the plugin as it is in this PR. Had a job using `withCredentials` that sleeps for 100 seconds using latest version of Credentials Binding. While the job is running I upgrade the plugin manually to the one in this PR, restart Jenkins instance and all is good. Job finishes successfully but it should not, right? Did I misunderstand something?

armfergom: Actually, I have anyway pushed what I think would fix the issue. When I am able to reproduce it I'd like to write a test for it...

jglick: Note that you could use just `charsetName` rather than `Charset.forName(charsetName)`.

pom.xml
```diff
@@ -17,16 +16,20 @@
   <description>Allows credentials to be bound to environment variables for use from miscellaneous build steps.
   </description>
   <url>http://wiki.jenkins-ci.org/display/JENKINS/Credentials+Binding+Plugin</url>
+  
   <properties>
     <workflow.version>1.7</workflow.version>
+    <jenkins.version>1.596.1</jenkins.version>
+    <java.level>6</java.level>
   </properties>
+
```
jglick: Please avoid trailing whitespace.

src/main/java/org/jenkinsci/plugins/credentialsbinding/Binding.java
```diff
@@ -25,12 +25,16 @@
...
 import com.cloudbees.plugins.credentials.common.StandardCredentials;
+
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.FilePath;
 import hudson.Launcher;
 import hudson.Util;
 import hudson.model.AbstractBuild;
 import hudson.model.BuildListener;
 import java.io.IOException;
+import java.io.Serializable;
```
jglick: :ant: unused import

jglick:
:bee:

armfergom:
Addressed your comments (unused imports, trailing whitespaces, and other suggestions) and squashed commits.

As for the potential issue with `charsetName` field, I agree we must cover that corner case, but it seems to me that `decorateLogger` (where `charsetName` is used) is only called when the [job is started](https://github.com/jenkinsci/credentials-binding-plugin/blob/master/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java#L95) but not ever again (or if it is resumed after a restart). I might be wrong, but I think that is why I was not able to reproduce the issue.

Thanks for the review.

jglick:
> and squashed commits

I would prefer you _not_ make destructive history operations in a pull request. Among other things, it makes it impossible to see what you changed since my last review.

jglick:
`decorateLogger` is called when the _step_ is started. The _build_ may have started days earlier, with several Jenkins restarts intervening.

Anyway the issue I hinted at would arise when you start the step with the old plugin version, then restart Jenkins with the new plugin version.

jglick:
re:bee:

PR #20, opened 2016-08-19
[JENKINS-37541] Protect ourselves against deserialization with null binding

Vlatombe:
Also bumped version of parent POM

https://issues.jenkins-ci.org/browse/JENKINS-37541
@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.9</version>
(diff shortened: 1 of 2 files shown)


jglick:
Hmm. :bee:, though I would like to know why this is happening, as it indicates a bug elsewhere.

Vlatombe:
Indeed, not sure what is the root cause (build record corruption maybe?). But in any case it never hurts to be defensive.

mattstep:
🐝

Vlatombe:
@reviewbybees done

PR #21, opened 2016-08-31
[JENKINS-37871] Add ordinal 100 to Extension annotation

hawky-4s-:
- so the credentials-binding plugin is executed at the beginning of the
  BuildWrapper phase. Otherwise it is executed too late, eg. when using
  the pre-release build steps of the release-plugin and no credentials
  are injected beforehand.

fixes [JENKINS-37871](https://issues.jenkins-ci.org/browse/JENKINS-37871)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
@@ -34,24 +34,16 @@ import hudson.model.BuildListener;
 import hudson.model.Run;
 import hudson.tasks.BuildWrapper;
 import hudson.tasks.BuildWrapperDescriptor;
-import java.io.IOException;
-import java.io.OutputStream;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.WeakHashMap;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
 import org.jenkinsci.plugins.credentialsbinding.MultiBinding;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


jglick:
Looks fine so far as it goes, but would prefer to have a test demonstrating the bug and proving the fix—for example a `@TestExtension` `BuildWrapperDescriptor` which verifies that some environment variable contains a pseudosecret.

hawky-4s-:
@jglick : I finally added a test for it. Unfortunately the CI tests are failing due to infrastructure issues. Could you retest it?

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -34,24 +34,16 @@
...
-import java.util.regex.Pattern;
-
 import org.jenkinsci.plugins.credentialsbinding.MultiBinding;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.*;
```
jglick: BTW please avoid star imports.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BuildWrapperOrderCredentialsBindingTest.java
```diff
@@ -0,0 +1,122 @@
...
+            public BuildWrapperOrderDescriptor() {
+                super(BuildWrapperOrder.class);
+            }
+
+            @Override
+            public boolean isApplicable(AbstractProject<?, ?> item) {
+                return true;
+            }
+
+            @Override
```
jglick: This can be omitted.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BuildWrapperOrderCredentialsBindingTest.java
```diff
@@ -0,0 +1,122 @@
...
+                }
+            }
+
+            return new Environment() {};
+        }
+
+        @TestExtension
+        public static class BuildWrapperOrderDescriptor extends BuildWrapperDescriptor {
+
+            public BuildWrapperOrderDescriptor() {
```
jglick: This can be omitted.

jglick:
> Could you retest it?

You can do so for yourself by `merge`ing with upstream; or by closing the PR, waiting a minute, and reopening.

PR #18, opened 2016-06-24
[JENKINS-28399] Add binding for SSH private keys

matthauck:
[JENKINS-28399](https://issues.jenkins-ci.org/browse/JENKINS-28399)

PR #22, opened 2016-09-08
WIP: DockerServerCredentialsBinding (request for comments/discussion)

thomasgl-orange:
**This is not a PR! (yet)**

Hi there,

I've started implementing a binding class for Docker client certificates.  Now, I'm wondering, where should the code go?  I've thought I should ask before spending to much time in the wrong code repository...

**Options 1:** add optional dependency from credentials-binding to docker-commons, and the new binding impl in credentials-binding. Same as what is proposed for SSH keys in #18, and also what I did so far in my prototype code. 

**Options 2:** implement new bindings as optional extensions in the plugins which provides the corresponding credentials type (docker-commons, ssh-credentials), with optional dependencies from these plugins to credentials-binding. 

To me, option 2 sounds better in the long term. It would avoid having one plugin with many optional dependencies, if many credentials type were to appear in the future. It also seems close to what would be the obvious choice if credentials binding was simply a standard feature of the main credentials plugin. 
What do you think/prefer?  

Thanks,
Thomas.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,6 +78,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>docker-commons</artifactId>
+      <version>1.0</version>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-support</artifactId>
       <version>${workflow.version}</version>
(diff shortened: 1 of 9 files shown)


jglick:
I would suggest just having a hard (not `optional`) dependency on `credentials-binding` in “domain” plugins like `docker-commons`.

thomasgl-orange:
Thanks Jesse for your suggestion. I think you're right. I'll close this PR, replaced it by:

- jenkinsci/credentials-binding-plugin#25 (a little refactoring to make some code reusable for my Binding implementation)
- jenkinsci/docker-commons-plugin#54 (the new Binding implementation, using code introduced in the other PR)

PR #23, opened 2016-09-08
Mention secret text as part of pipeline option.

kwhetstone:
You can use the secret text credentials with the pipeline step.  This should be added to the documentation.  I'm copied this from [#347](https://github.com/jenkins-infra/jenkins.io/pull/347) since that file is generated from the help.html from this plugin.

cc @nzakas to ensure the info is still accurate


jglick:
The step help was only intended to show the overall usage, not mention every possible kind of binding, for which see _Snippet Generator_. But this was a common enough one.

PR #24, opened 2016-09-09
Adding symbols to binding types for easier usage

jglick:
@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,6 +65,11 @@
       <version>${workflow.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>structs</artifactId>
+      <version>1.5</version>
+    </dependency>
+    <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-aggregator</artifactId>
       <version>${workflow.version}</version>
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,8 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.14</version>
+    <version>2.17</version>
+    <relativePath/>
   </parent>
 
   <artifactId>credentials-binding</artifactId>
@@ -17,9 +18,7 @@
   </description>
   <url>http://wiki.jenkins-ci.org/display/JENKINS/Credentials+Binding+Plugin</url>
   <properties>
-    <workflow.version>1.7</workflow.version>
-    <jenkins.version>1.596.1</jenkins.version>
-    <java.level>6</java.level>
+    <jenkins.version>1.642.3</jenkins.version>
   </properties>  
   <licenses>
     <license>
@@ -38,13 +37,13 @@
   <repositories>
     <repository>
       <id>repo.jenkins-ci.org</id>
-      <url>http://repo.jenkins-ci.org/public/</url>
+      <url>https://repo.jenkins-ci.org/public/</url>
     </repository>
   </repositories>
   <pluginRepositories>
     <pluginRepository>
       <id>repo.jenkins-ci.org</id>
-      <url>http://repo.jenkins-ci.org/public/</url>
+      <url>https://repo.jenkins-ci.org/public/</url>
     </pluginRepository>
   </pluginRepositories>
 
@@ -52,17 +51,17 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>credentials</artifactId>
-      <version>1.23</version>
+      <version>2.1.7</version>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>plain-credentials</artifactId>
-      <version>1.0</version>
+      <version>1.3</version>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-step-api</artifactId>
-      <version>${workflow.version}</version>
+      <version>2.4</version>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
@@ -71,21 +70,46 @@
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-aggregator</artifactId>
-      <version>${workflow.version}</version>
+      <artifactId>workflow-cps</artifactId>
+      <version>2.21</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-job</artifactId>
+      <version>2.7</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-durable-task-step</artifactId>
+      <version>2.5</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-basic-steps</artifactId>
+      <version>2.2</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-cps</artifactId>
+      <version>2.21</version>
+      <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-step-api</artifactId>
-      <version>${workflow.version}</version>
+      <version>2.4</version>
       <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-support</artifactId>
-      <version>${workflow.version}</version>
+      <version>2.10</version>
       <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
@@ -172,8 +172,7 @@ public final class BindingStep extends AbstractStepImpl {
 
     }
 
-    // TODO use BodyExecutionCallback.TailCall from https://github.com/jenkinsci/workflow-plugin/pull/168
-    private static final class Callback extends BodyExecutionCallback {
+    private static final class Callback extends BodyExecutionCallback.TailCall {
 
         private static final long serialVersionUID = 1;
 
@@ -183,24 +182,22 @@ public final class BindingStep extends AbstractStepImpl {
             this.unbinders = unbinders;
         }
 
-        private void cleanup(StepContext context) {
+        @Override protected void finished(StepContext context) throws Exception {
+            Exception xx = null;
             for (MultiBinding.Unbinder unbinder : unbinders) {
                 try {
                     unbinder.unbind(context.get(Run.class), context.get(FilePath.class), context.get(Launcher.class), context.get(TaskListener.class));
                 } catch (Exception x) {
-                    context.onFailure(x);
+                    if (xx == null) {
+                        xx = x;
+                    } else {
+                        xx.addSuppressed(x);
+                    }
                 }
             }
-        }
-
-        @Override public void onSuccess(StepContext context, Object result) {
-            cleanup(context);
-            context.onSuccess(result);
-        }
-
-        @Override public void onFailure(StepContext context, Throwable t) {
-            cleanup(context);
-            context.onFailure(t);
+            if (xx != null) {
+                throw xx;
+            }
         }
 
     }
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/FileBinding.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/FileBinding.java
@@ -32,6 +32,7 @@ import hudson.model.TaskListener;
 import hudson.slaves.WorkspaceList;
 import java.io.IOException;
 import java.util.UUID;
+import org.jenkinsci.Symbol;
 
 import org.jenkinsci.plugins.credentialsbinding.Binding;
 import org.jenkinsci.plugins.credentialsbinding.BindingDescriptor;
@@ -97,6 +98,7 @@ public class FileBinding extends Binding<FileCredentials> {
         secret.copyFrom(credentials.getContent());
     }
 
+    @Symbol("file")
     @Extension public static class DescriptorImpl extends BindingDescriptor<FileCredentials> {
 
         @Override protected Class<FileCredentials> type() {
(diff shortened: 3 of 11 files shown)


src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep/help.html
```diff
@@ -3,20 +3,29 @@
...
+  withCredentials([usernameColonPassword(credentialsId: 'mylogin', variable: 'USERPASS')]) {
     sh '''
       set +x
-      curl -u $USERNAME:$PASSWORD https://private.server/ > output
+      curl -u $USERPASS https://private.server/ > output
+    '''
+  }
+}</code></pre>
+<p>
+    As another example (use <i>Snippet Generator</i> to see all options):
```
jglick: Reworked #23.

src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep/help.html
```diff
@@ -3,20 +3,29 @@
     Each binding will define an environment variable active within the scope of the step.
     You can then use them directly from any other steps that expect environment variables to be set:
 </p>
-<pre><code>
-node {
-  withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: 'mylogin',
-                    usernameVariable: 'USERNAME', passwordVariable: 'PASSWORD']]) {
+<pre><code>node {
+  withCredentials([usernameColonPassword(credentialsId: 'mylogin', variable: 'USERPASS')]) {
```
jglick: Of course only works with sufficiently new Pipeline plugins, but I will assume that is most people going forward. _Snippet Generator_ will always produce syntax compatible with what you actually have.

abayer:
:bee: Symbolize all the things!

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/StringBinding.java
```diff
@@ -50,6 +51,7 @@
         return new SingleEnvironment(getCredentials(build).getSecret().getPlainText());
     }
 
+    @Symbol("string")
```
svanoort: :bug:  Does "string" immediately say credentials to users?  I feel like this will bite us -- maybe stringCredential instead?  Or stringCredentialBinding?

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/UsernamePasswordMultiBinding.java
```diff
@@ -76,6 +77,7 @@ public String getPasswordVariable() {
         return new HashSet<String>(Arrays.asList(usernameVariable, passwordVariable));
     }
 
+    @Symbol("usernamePassword")
```
svanoort: I liked it.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBinding.java
```diff
@@ -53,6 +54,7 @@
         secret.unzipFrom(credentials.getContent());
     }
 
+    @Symbol("zip")
```
svanoort: 🐛  I think - didn't we already have a zip step somewhere?

Maybe zipCredential or zipSecret?

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/FileBinding.java
```diff
@@ -97,6 +98,7 @@ protected void copy(FilePath secret, FileCredentials credentials) throws IOExcep
         secret.copyFrom(credentials.getContent());
     }
 
+    @Symbol("file")
```
svanoort: I think 🐛 - as with the others for namespace collision with other things - probably fileCredential or fileSecret is much easier to understand than just 'file'

jglick: Namespace collision with that? Symbols are scoped per extension point. What other `MultiBinding` would want to be called `file`?

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/StringBinding.java
jglick: Look at the example.

```groovy
withCredentials([string(…)]) {…}
```

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBinding.java
jglick: I do not know of a `zip` step offhand. If there is one, _Snippet Generator_ will automatically avoid using the symbol.

svanoort: Pipeline utility steps has one - https://github.com/jenkinsci/pipeline-utility-steps-plugin/blob/master/docs/STEPS.md

In general, would suggest using slightly more descriptive name.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/FileBinding.java
svanoort: Ah, okay I didn't realize the scoping role was per-extension.  But I still think calling it 'file' is less helpful than fileCredential (which is what it is, as opposed to things which read a file directly).

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/StringBinding.java
svanoort: I had seen that already, yes -- I'll downgrade it an 🐜  because collisions aren't as bad as I thought, but think it's going to be immortalized in pipeline scripts better to be a little more descriptive.

jglick:
Regarding shorter vs. longer names, see the [Javadoc](https://github.com/jenkinsci/structs-plugin/blob/ad2ff9a7db4ff03e1533503b6ee1fee6b0a0c5c4/annotation/src/main/java/org/jenkinsci/Symbol.java#L21-L23): including something like “credentials” or “secret” in the name is redundant given the context of the calling step.

A clash with a `zip` step (if you install that plugin) is potentially a concern; steps take precedence over symbols, so when using that plugin, you would need to fall back to the old syntax with this PR as written. `Symbol` does permit variants, but I need to verify that `DSL` accepts any variant, and that `Snippetizer` will choose the first unambiguous variant it finds.

It might also be useful to useful to introduce a `symbols` global variable to disambiguate without falling back to `$class` maps, by analogy with the longstanding (if rarely used) `steps` variable.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBindingTest.java
```diff
@@ -23,64 +23,35 @@
...
-            FileItem fi = new FileItemImpl(zip);
-            FileCredentialsImpl fc = new FileCredentialsImpl(CredentialsScope.GLOBAL, credentialsId, "Just a zip file", fi, fi.getName(), null);
-            CredentialsProvider.lookupStores(j.jenkins).iterator().next().addCredentials(Domain.global(), fc);
-        }
-        finally {
-            IOUtils.closeQuietly(zipStream);
-            zipStream = null;
-        }
-
+        FileCredentialsImpl fc = new FileCredentialsImpl(CredentialsScope.GLOBAL, credentialsId, "Just a zip file", "a.zip", SecretBytes.fromBytes(IOUtils.toByteArray(ZipFileBindingTest.class.getResource("a.zip"))));
```
jglick: Simplified thanks to @stephenc’s https://github.com/jenkinsci/plain-credentials-plugin/pull/6.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -172,8 +172,7 @@ private Object readResolve() throws ObjectStreamException {
 
     }
 
-    // TODO use BodyExecutionCallback.TailCall from https://github.com/jenkinsci/workflow-plugin/pull/168
-    private static final class Callback extends BodyExecutionCallback {
+    private static final class Callback extends BodyExecutionCallback.TailCall {
```
jglick: Made possible by new dep.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/UsernamePasswordBindingTest.java
```diff
@@ -58,7 +57,7 @@
         FreeStyleProject p = r.createFreeStyleProject();
         p.getBuildWrappersList().add(new SecretBuildWrapper(Collections.<Binding<?>>singletonList(new UsernamePasswordBinding("AUTH", c.getId()))));
         p.getBuildersList().add(new Shell("set +x\necho $AUTH > auth.txt"));
-        r.configRoundtrip((Item)p);
+        r.configRoundtrip(p);
```
jglick: Newer `jenkins-test-harness`.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
```diff
@@ -194,10 +212,7 @@
                 assertNotNull(b);
                 assertEquals(Collections.<String>emptySet(), grep(b.getRootDir(), secret));
                 SemaphoreStep.success("cleanupAfterRestart/1", null);
-                while (b.isBuilding()) { // TODO 1.607+ use waitForCompletion
-                    Thread.sleep(100);
-                }
-                story.j.assertBuildStatusSuccess(b);
+                story.j.assertBuildStatusSuccess(story.j.waitForCompletion(b));
```
jglick: Newer `jenkins-test-harness`.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
```diff
@@ -168,16 +188,14 @@
         final String secret = "s3cr3t";
         story.addStep(new Statement() {
             @Override public void evaluate() throws Throwable {
-                File originalSecret = tmp.newFile();
-                FileUtils.write(originalSecret, secret);
-                FileCredentialsImpl c = new FileCredentialsImpl(CredentialsScope.GLOBAL, "creds", "sample", new FileParameterValue.FileItemImpl(originalSecret), null, null);
+                FileCredentialsImpl c = new FileCredentialsImpl(CredentialsScope.GLOBAL, "creds", "sample", "secret.txt", SecretBytes.fromBytes(secret.getBytes()));
```
jglick: Cf. `ZipFileBindingTest`.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
```diff
@@ -108,7 +128,7 @@
                 WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
                 p.setDefinition(new CpsFlowDefinition(""
                         + "node {\n"
-                        + "  withCredentials([[$class: 'UsernamePasswordMultiBinding', usernameVariable: 'USERNAME', passwordVariable: 'PASSWORD', credentialsId: '" + credentialsId + "']]) {\n"
+                        + "  withCredentials([usernamePassword(usernameVariable: 'USERNAME', passwordVariable: 'PASSWORD', credentialsId: '" + credentialsId + "')]) {\n"
```
jglick: With a sufficiently new `workflow-cps` test dep, we can verify the new symbols in context.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
```diff
@@ -93,9 +99,23 @@
...
+                new SnippetizerTester(story.j).assertRoundTrip(new BindingStep(Collections.<MultiBinding>singletonList(new ZipFileBinding("file", "secrets"))),
+                    "withCredentials([[$class: 'ZipFileBinding', credentialsId: 'secrets', variable: 'file']]) {\n    // some block\n}");
             }
         });
     }
+    public static class ZipStep extends AbstractStepImpl {
+        @DataBoundConstructor public ZipStep() {}
+        @TestExtension("configRoundTrip") public static class DescriptorImpl extends AbstractStepDescriptorImpl {
+            public DescriptorImpl() {super(Execution.class);}
+            @Override public String getFunctionName() {return "zip";}
```
jglick: Example of a clashing step name.

pom.xml
```diff
@@ -17,9 +18,7 @@
   </description>
   <url>http://wiki.jenkins-ci.org/display/JENKINS/Credentials+Binding+Plugin</url>
   <properties>
-    <workflow.version>1.7</workflow.version>
-    <jenkins.version>1.596.1</jenkins.version>
-    <java.level>6</java.level>
+    <jenkins.version>1.642.3</jenkins.version>
```
jglick: Required in order to test against recent `workflow-cps` releases.

pom.xml
```diff
@@ -38,49 +37,79 @@
...
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>credentials</artifactId>
-      <version>1.23</version>
+      <version>2.1.7</version>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>plain-credentials</artifactId>
-      <version>1.0</version>
+      <version>1.3</version>
```
jglick: Better `FileCredentialsImpl` signature.

pom.xml
```diff
@@ -38,49 +37,79 @@
...
+      <url>https://repo.jenkins-ci.org/public/</url>
     </pluginRepository>
   </pluginRepositories>
 
   <dependencies>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>credentials</artifactId>
-      <version>1.23</version>
+      <version>2.1.7</version>
```
jglick: Needed by `plain-credentials` update.

pom.xml
```diff
@@ -38,49 +37,79 @@
...
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-step-api</artifactId>
-      <version>${workflow.version}</version>
+      <version>2.4</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>structs</artifactId>
```
jglick: Needed for `@Symbol`.

pom.xml
```diff
@@ -38,49 +37,79 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-durable-task-step</artifactId>
+      <version>2.5</version>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-aggregator</artifactId>
```
jglick: Test deps on `workflow-aggregator` are a bad idea generally.

pom.xml
```diff
@@ -38,49 +37,79 @@
...
-      <version>${workflow.version}</version>
+      <artifactId>workflow-basic-steps</artifactId>
+      <version>2.2</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-cps</artifactId>
+      <version>2.21</version>
+      <classifier>tests</classifier>
```
jglick: `SnippetizerTester`

pom.xml
```diff
@@ -17,9 +18,7 @@
   </description>
   <url>http://wiki.jenkins-ci.org/display/JENKINS/Credentials+Binding+Plugin</url>
   <properties>
-    <workflow.version>1.7</workflow.version>
```
jglick: Meaningless as of 2.x.

abayer:
:bee:

jglick:
Waiting for @svanoort to recheck I guess.

svanoort:
Somewhat blind re-:bee:

PR #27, opened 2016-10-18
[JENKINS-38831] Add support for credential tracking in SecretBuildWrapper and BindingStep

iwarapter:
[JENKINS-38831](https://issues.jenkins-ci.org/browse/JENKINS-38831)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
@@ -30,7 +30,10 @@ import com.cloudbees.plugins.credentials.common.StandardCredentials;
 import hudson.ExtensionPoint;
 import hudson.FilePath;
 import hudson.Launcher;
-import hudson.model.*;
+import hudson.model.AbstractDescribableImpl;
+import hudson.model.Descriptor;
+import hudson.model.Run;
+import hudson.model.TaskListener;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.Serializable;
@@ -116,12 +119,13 @@ public abstract class MultiBinding<C extends StandardCredentials> extends Abstra
      */
     protected final @Nonnull C getCredentials(@Nonnull Run<?,?> build) throws IOException {
         IdCredentials cred = CredentialsProvider.findCredentialById(credentialsId, IdCredentials.class, build);
-        CredentialsProvider.track(build, cred);
         if (cred==null)
             throw new CredentialNotFoundException(credentialsId);
 
-        if (type().isInstance(cred))
+        if (type().isInstance(cred)) {
+            CredentialsProvider.track(build, cred);
             return type().cast(cred);
+        }
 
         
         Descriptor expected = Jenkins.getActiveInstance().getDescriptor(type());
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
@@ -24,17 +24,13 @@
 
 package org.jenkinsci.plugins.credentialsbinding;
 
-import com.cloudbees.plugins.credentials.CredentialsDescriptor;
 import com.cloudbees.plugins.credentials.CredentialsProvider;
 import com.cloudbees.plugins.credentials.common.IdCredentials;
 import com.cloudbees.plugins.credentials.common.StandardCredentials;
 import hudson.ExtensionPoint;
 import hudson.FilePath;
 import hudson.Launcher;
-import hudson.model.AbstractDescribableImpl;
-import hudson.model.Descriptor;
-import hudson.model.Run;
-import hudson.model.TaskListener;
+import hudson.model.*;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.Serializable;
@@ -62,7 +58,7 @@ public abstract class MultiBinding<C extends StandardCredentials> extends Abstra
     }
 
     /** Type token. */
-    public abstract Class<C> type();
+    protected abstract Class<C> type();
 
     /** Identifier of the credentials to be bound. */
     public final String getCredentialsId() {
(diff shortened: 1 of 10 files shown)


iwarapter:
This is currently setup to track the instances of builds using the credential, would we want this to be that granular or do we just want the Job?

src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
```diff
@@ -62,7 +62,7 @@ protected MultiBinding(String credentialsId) {
     }
 
     /** Type token. */
-    protected abstract Class<C> type();
+    public abstract Class<C> type();
```
jglick: Binary-incompatible for external implementations, I think.

Probably solvable by adding a `public final` method to `MultiBinding` to set up the tracking.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -92,6 +94,8 @@
                 MultiBinding.MultiEnvironment environment = binding.bind(run, workspace, launcher, listener);
                 unbinders.add(environment.getUnbinder());
                 overrides.putAll(environment.getValues());
+                IdCredentials id = CredentialsProvider.findCredentialById(binding.getCredentialsId(), binding.type(), run);
+                CredentialsProvider.track(run, id);
```
jglick: Perhaps also desirable to call `track(Node, Credentials)`. (For `BindingStep`, just inject a `Node`; for `SecretBuildWrapper`, use the method on `AbstractBuild` IIRC.) @stephenc any guidance?

stephenc: So track(Node, C) is to track when the credential is used *to launch that node*. Not the case here. We already have tracking between the Run and the Node(s) it ran on.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -92,6 +94,8 @@
                 MultiBinding.MultiEnvironment environment = binding.bind(run, workspace, launcher, listener);
                 unbinders.add(environment.getUnbinder());
                 overrides.putAll(environment.getValues());
+                IdCredentials id = CredentialsProvider.findCredentialById(binding.getCredentialsId(), binding.type(), run);
```
stephenc: Ugh! The tracking should be done when the credential is actually looked up... not by doing a second look up... as that second look-up may trigger a second logged call to an external credentials management system. Much better if we can have it as part of the binding... May require a thread local to capture the Run context

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -57,6 +59,8 @@
         final List<MultiBinding.MultiEnvironment> m = new ArrayList<MultiBinding.MultiEnvironment>();
         for (MultiBinding binding : bindings) {
             m.add(binding.bind(build, build.getWorkspace(), launcher, listener));
+            IdCredentials id = CredentialsProvider.findCredentialById(binding.getCredentialsId(), binding.type(), build);
```
stephenc: Ugh! Please no... one lookup only

iwarapter:
@jglick @stephenc thanks for your feedback, this should fix all the issues i believe.

iwarapter:
The jenkins hook seems to be stuck, can someone give it a kick :)

iwarapter:
Is there anything else needed for this PR?

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
jglick: Well, not in the case of Pipeline builds generally, but I see your point—`track(Node, Credentials)` is irrelevant here. I will file a PR to fix the Javadoc.

src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
```diff
@@ -24,17 +24,13 @@
...
 import com.cloudbees.plugins.credentials.common.IdCredentials;
 import com.cloudbees.plugins.credentials.common.StandardCredentials;
 import hudson.ExtensionPoint;
 import hudson.FilePath;
 import hudson.Launcher;
-import hudson.model.AbstractDescribableImpl;
-import hudson.model.Descriptor;
-import hudson.model.Run;
-import hudson.model.TaskListener;
+import hudson.model.*;
```
jglick: Please revert and use explicit imports.

src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
```diff
@@ -120,6 +116,7 @@ public Unbinder getUnbinder() {
      */
     protected final @Nonnull C getCredentials(@Nonnull Run<?,?> build) throws IOException {
         IdCredentials cred = CredentialsProvider.findCredentialById(credentialsId, IdCredentials.class, build);
+        CredentialsProvider.track(build, cred);
```
jglick: Better moved to the `if` block that `return`s a value.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/UsernamePasswordBindingTest.java
```diff
@@ -24,31 +24,39 @@
 
 package org.jenkinsci.plugins.credentialsbinding.impl;
 
-import com.cloudbees.plugins.credentials.CredentialsProvider;
-import com.cloudbees.plugins.credentials.CredentialsScope;
+import com.cloudbees.plugins.credentials.*;
```
jglick: Ditto, please retain explicit imports.

iwarapter:
@jglick updated from feedback.

PR #28, opened 2016-10-31
[JENKINS-24805] Start masking secrets in freestyle logs.

abayer:
[JENKINS-24805](https://issues.jenkins-ci.org/browse/JENKINS-24805)

This still needs tests - I just want to make sure the approach is right.

cc @reviewbybees esp @jglick

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
@@ -38,12 +38,15 @@ import hudson.model.TaskListener;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.Serializable;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 import javax.annotation.Nonnull;
 
+import hudson.util.Secret;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.credentialsbinding.impl.CredentialNotFoundException;
 import org.kohsuke.stapler.DataBoundConstructor;
@@ -138,4 +141,19 @@ public abstract class MultiBinding<C extends StandardCredentials> extends Abstra
         return (BindingDescriptor<C>) super.getDescriptor();
     }
 
+    /**
+     * Utility method for turning a collection of secret strings into a {@link Secret}.
+     * @param secrets A collection of secret strings
+     * @return A {@link Secret} generated from that collection.
+     */
+    public static Secret getSecretForStrings(Collection<String> secrets) {
+        StringBuilder b = new StringBuilder();
+        for (String secret : secrets) {
+            if (b.length() > 0) {
+                b.append('|');
+            }
+            b.append(Pattern.quote(secret));
+        }
+        return Secret.fromString(b.toString());
+    }
 }
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
@@ -49,6 +49,8 @@ import javax.annotation.Nonnull;
 import hudson.util.Secret;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.credentialsbinding.impl.CredentialNotFoundException;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 /**
@@ -142,11 +144,12 @@ public abstract class MultiBinding<C extends StandardCredentials> extends Abstra
     }
 
     /**
-     * Utility method for turning a collection of secret strings into a {@link Secret}.
+     * Utility method for turning a collection of secret strings into a single {@link String} for pattern compilation.
      * @param secrets A collection of secret strings
-     * @return A {@link Secret} generated from that collection.
+     * @return A {@link String} generated from that collection.
      */
-    public static Secret getSecretForStrings(Collection<String> secrets) {
+    @Restricted(NoExternalUse.class)
+    public static String getPatternStringForSecrets(Collection<String> secrets) {
         StringBuilder b = new StringBuilder();
         for (String secret : secrets) {
             if (b.length() > 0) {
@@ -154,6 +157,6 @@ public abstract class MultiBinding<C extends StandardCredentials> extends Abstra
             }
             b.append(Pattern.quote(secret));
         }
-        return Secret.fromString(b.toString());
+        return b.toString();
     }
 }
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
@@ -98,7 +98,9 @@ public class SecretBuildWrapper extends BuildWrapper {
             secrets.addAll(e.getValues().values());
         }
 
-        secretsForBuild.put(build, secrets);
+        if (!secrets.isEmpty()) {
+            secretsForBuild.put(build, secrets);
+        }
 
         return new Environment() {
             @Override public void buildEnvVars(Map<String,String> env) {
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
@@ -35,9 +35,7 @@ import hudson.model.Run;
 import hudson.tasks.BuildWrapper;
 import hudson.tasks.BuildWrapperDescriptor;
 import java.io.IOException;
-import java.io.ObjectStreamException;
 import java.io.OutputStream;
-import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -70,7 +68,7 @@ public class SecretBuildWrapper extends BuildWrapper {
      */
     public static @CheckForNull Pattern getPatternForBuild(@Nonnull AbstractBuild<?, ?> build) {
         if (secretsForBuild.containsKey(build)) {
-            return Pattern.compile(MultiBinding.getSecretForStrings(secretsForBuild.get(build)).getPlainText());
+            return Pattern.compile(MultiBinding.getPatternStringForSecrets(secretsForBuild.get(build)));
         } else {
             return null;
         }
@@ -86,11 +84,7 @@ public class SecretBuildWrapper extends BuildWrapper {
 
     @Override
     public OutputStream decorateLogger(AbstractBuild build, OutputStream logger) throws IOException, InterruptedException, Run.RunnerAbortedException {
-        if (!bindings.isEmpty()) {
-            return new Filter(build.getCharset().name()).decorateLogger(build, logger);
-        } else {
-            return logger;
-        }
+        return new Filter(build.getCharset().name()).decorateLogger(build, logger);
     }
 
     @Override public Environment setUp(AbstractBuild build, final Launcher launcher, BuildListener listener) throws IOException, InterruptedException {
@@ -104,9 +98,7 @@ public class SecretBuildWrapper extends BuildWrapper {
             secrets.addAll(e.getValues().values());
         }
 
-        if (!secrets.isEmpty()) {
-            secretsForBuild.put(build, secrets);
-        }
+        secretsForBuild.put(build, secrets);
 
         return new Environment() {
             @Override public void buildEnvVars(Map<String,String> env) {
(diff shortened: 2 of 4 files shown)


src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
```diff
@@ -138,4 +141,19 @@ public Unbinder getUnbinder() {
...
+     */
+    public static Secret getSecretForStrings(Collection<String> secrets) {
+        StringBuilder b = new StringBuilder();
+        for (String secret : secrets) {
+            if (b.length() > 0) {
+                b.append('|');
+            }
+            b.append(Pattern.quote(secret));
+        }
+        return Secret.fromString(b.toString());
```
jglick: No need to return `Secret` here; return `Pattern`. It is only `BindingStep` which cares about wrapping that in a `Secret`, for serialization purposes.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -26,48 +26,95 @@
...
 
+    private static Map<AbstractBuild<?, ?>, Collection<String>> secretsForBuild = new WeakHashMap<AbstractBuild<?, ?>, Collection<String>>();
+
+    /**
+     * Gets the {@link Pattern} for the secret values for a given build, if that build has secrets defined. If not, return
+     * null.
+     * @param build A non-null build.
+     * @return A compiled {@link Pattern} from the build's secret values, if the build has any.
+     */
+    public static @Nullable Pattern getPatternForBuild(@Nonnull AbstractBuild<?, ?> build) {
```
jglick: Use `@CheckForNull`.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -26,48 +26,95 @@
...
 
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
 @SuppressWarnings({"rawtypes", "unchecked"}) // inherited from BuildWrapper
 public class SecretBuildWrapper extends BuildWrapper {
 
     private /*almost final*/ List<? extends MultiBinding<?>> bindings;
 
+    private static Map<AbstractBuild<?, ?>, Collection<String>> secretsForBuild = new WeakHashMap<AbstractBuild<?, ?>, Collection<String>>();
```
jglick: `final`

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -26,48 +26,95 @@
...
             m.add(binding.bind(build, build.getWorkspace(), launcher, listener));
         }
+
+        secretsForBuild.put(build, new HashSet<String>());
+
         return new Environment() {
             @Override public void buildEnvVars(Map<String,String> env) {
                 for (MultiBinding.MultiEnvironment e : m) {
                     env.putAll(e.getValues());
+                    secretsForBuild.get(build).addAll(e.getValues().values());
```
jglick: Better to add this above, as soon as we have called `bind`, so we do not have to assume `buildEnvVars` has been called.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -86,6 +133,52 @@ protected Object readResolve() {
         return this;
     }
 
+    /** Similar to {@code MaskPasswordsOutputStream}. */
+    private static final class Filter extends ConsoleLogFilter implements Serializable {
```
jglick: Need not be `Serializable`.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -86,6 +133,52 @@ protected Object readResolve() {
...
+
+        private static final long serialVersionUID = 1;
```
jglick: Ditto.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -86,6 +133,52 @@ protected Object readResolve() {
...
+            this.charsetName = charsetName;
+        }
+
+        // To avoid de-serialization issues with newly added field (charsetName)
+        private Object readResolve() throws ObjectStreamException {
+            if (this.charsetName == null) {
+                this.charsetName = Charsets.UTF_8.name();
+            }
+            return this;
+        }
```
jglick: Unnecessary, delete.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -86,6 +133,52 @@ protected Object readResolve() {
         return this;
     }
 
+    /** Similar to {@code MaskPasswordsOutputStream}. */
+    private static final class Filter extends ConsoleLogFilter implements Serializable {
+
+        private static final long serialVersionUID = 1;
+
+        private String charsetName;
```
jglick: `final`

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
abayer: Ah, ok, you'd mentioned something about doing it here.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
abayer: Whoops. Done.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
abayer: Done.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
abayer: Done.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
abayer: Done.

src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
jglick: And only `BindingStep` needs it to be a `Secret`. `SecretBuildWrapper` just throws that away and recompiles the `Pattern`—a waste, and confusing.

src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
```diff
@@ -138,4 +141,19 @@ public Unbinder getUnbinder() {
         return (BindingDescriptor<C>) super.getDescriptor();
     }
 
+    /**
+     * Utility method for turning a collection of secret strings into a {@link Secret}.
+     * @param secrets A collection of secret strings
+     * @return A {@link Secret} generated from that collection.
+     */
+    public static Secret getSecretForStrings(Collection<String> secrets) {
```
jglick: Best to make this `@Restricted(NoExternalUse.class)`, or move it somewhere in the `impl` package and make it package access.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -26,38 +26,88 @@
...
     public List<? extends MultiBinding<?>> getBindings() {
         return bindings;
     }
 
+    @Override
+    public OutputStream decorateLogger(AbstractBuild build, OutputStream logger) throws IOException, InterruptedException, Run.RunnerAbortedException {
+        if (!bindings.isEmpty()) {
+            return new Filter(build.getCharset().name()).decorateLogger(build, logger);
+        } else {
+            return logger;
```
jglick: I guess…there is no use case for adding the wrapper with an empty list of bindings, though, so we could just skip this check.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
```diff
@@ -26,38 +26,88 @@
...
+        Set<String> secrets = new HashSet<String>();
+
         for (MultiBinding binding : bindings) {
-            m.add(binding.bind(build, build.getWorkspace(), launcher, listener));
+            MultiBinding.MultiEnvironment e = binding.bind(build, build.getWorkspace(), launcher, listener);
+            m.add(e);
+            secrets.addAll(e.getValues().values());
+        }
+
+        if (!secrets.isEmpty()) {
```
jglick: Simpler to just add it unconditionally.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapperTest.java
```diff
@@ -0,0 +1,81 @@
...
+        SecretBuildWrapper bw_1 = new SecretBuildWrapper(Collections.<MultiBinding<?>>singletonList(new StringBinding("PASS_1", credentialsId_1)));
+        SecretBuildWrapper bw_2 = new SecretBuildWrapper(Collections.<MultiBinding<?>>singletonList(new StringBinding("PASS_2", credentialsId_2)));
+
+        FreeStyleProject f = r.createFreeStyleProject();
+
+        f.setConcurrentBuild(true);
+        f.getBuildersList().add(new Shell("echo $PASS_1"));
+        f.getBuildersList().add(new Shell("echo $PASS_2"));
+        f.getBuildWrappersList().add(bw_1);
+        f.getBuildWrappersList().add(bw_2);
```
jglick: Not very realistic, since you can in fact only add one of each kind of `BuildWrapper` to a given `Project`. Better to create just one wrapper, with a two-element list of bindings.

src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
abayer: Done.

src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
abayer: Gotcha - fixing up.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
abayer: Fixing.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapperTest.java
abayer: Actually, removed the second one entirely since that's a separate issue.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
jglick: I suppose it would just work as a degenerate case without special effort, but if you are worried you should test it.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapperTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+        SecretBuildWrapper wrapper = new SecretBuildWrapper(Collections.singletonList(new StringBinding("PASS_1", firstCredentialsId)));
+
+        FreeStyleProject f = r.createFreeStyleProject();
+
+        f.setConcurrentBuild(true);
+        f.getBuildersList().add(new Shell("echo $PASS_1"));
+        f.getBuildWrappersList().add(wrapper);
+
+        r.configRoundtrip((Item)f);
```
jglick: BTW the cast is unnecessary as of some sufficiently recent `jenkins-test-harness` ~ `plugin-pom`.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SecretBuildWrapper.java
abayer: Good point - turned out to be busted, so I've got a new commit with test+fix incoming.

aheritier:
LGTM 🐝

PR #31, opened 2016-11-21
[JENKINS-39906] Fix PCT against core >= 2.16

andresrc:
See [JENKINS-39906](https://issues.jenkins-ci.org/browse/JENKINS-39906)

Fixed by picking needed test dependencies. Upgrade to newer parent POM added as a nice-to-have.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,12 +70,6 @@
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
-      <artifactId>subversion</artifactId>
-      <version>2.5</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -69,6 +69,24 @@
       <version>1.5</version>
     </dependency>
     <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>subversion</artifactId>
+      <version>2.5</version>
```
jglick: Is this to override some 1.x bundled dependency? I thought that would be obsolete as of core 2.x.

If this is because of a dependency from `workflow-scm-step`, then make sure _that_ is updated (newer versions do not depend on `subversion`), and remove this dependency.

andresrc: Thanks, that did the trick, new commit added.

jglick:
Is this now obsolete?

PR #33, opened 2017-02-06
[JENKINS-41752] Fix PCT for 2.32.2

varyvol:
[JENKINS-41752](https://issues.jenkins-ci.org/browse/JENKINS-41752)

- Force the creation of the user, to avoid depending on new User.ALLOW_USER_CREATION_VIA_URL property.
- Update test dependencies for instance to start.

@reviewbybees @stephenc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -132,7 +132,7 @@
       <version>1.1.0</version>
       <scope>test</scope>
     </dependency>
-        <dependency>
+    <dependency>
       <groupId>org.xmlunit</groupId>
       <artifactId>xmlunit-matchers</artifactId>
       <version>2.2.0</version>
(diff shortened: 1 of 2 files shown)


armfergom:
🐝 as far as I understand it

raul-arabaolaza:
Does [this comment](https://github.com/jenkinsci/credentials-binding-plugin/pull/31#discussion_r88903758) apply here as well? I mean, do we really need the subversion dependency  or can be avoided?

varyvol:
@raul-arabaolaza it did. Anyway I think at the end is just a necessary dependency for the tests, but probably this way it's "cleaner". Changed.

raul-arabaolaza:
I prefer this way, is coherent with [#PR-31](https://github.com/jenkinsci/credentials-binding-plugin/pull/31) and we do not add unneeded dependencies (the less subversion added the merrier :)) 

:bee:

PR #34, opened 2017-03-21
[FIXED JENKINS-42999] Allow bindings to specify their required context

abayer:
[JENKINS-42999](https://issues.jenkins-ci.org/browse/JENKINS-42999)

Not all bindings require a workspace, and those that don't should be
able to be used in `withCredentials` outside of a `node` block. This
adds `BindingDescriptor.getRequiredContext()`, defaulting to the four
contexts that were previously required by `BindingStep`, but allowing
override. `BindingStep` will now throw a
`MissingContextVariableException` if any of the `MultiBinding`s used
have a required context that cannot be satisfied, as well as the
normal potential `MissingContextVariableException` for `BindingStep.DescriptorImpl.getRequiredContext()`.

In addition, bumped `workflow-step-api` to 2.9 and moved `BindingStep`
and friends to non-deprecated code paths while I was here.

cc @reviewbybees esp @jglick @stephenc @rsandell

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/BindingDescriptor.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/BindingDescriptor.java
@@ -29,11 +29,21 @@ import com.cloudbees.plugins.credentials.CredentialsProvider;
 import com.cloudbees.plugins.credentials.common.AbstractIdCredentialsListBoxModel;
 import com.cloudbees.plugins.credentials.common.StandardCredentials;
 import com.cloudbees.plugins.credentials.domains.DomainRequirement;
+import hudson.FilePath;
+import hudson.Launcher;
 import hudson.model.Descriptor;
 import hudson.model.Item;
+import hudson.model.Run;
+import hudson.model.TaskListener;
 import hudson.security.ACL;
 import hudson.util.ListBoxModel;
+
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.kohsuke.stapler.AncestorInPath;
 
 /**
@@ -43,6 +53,19 @@ import org.kohsuke.stapler.AncestorInPath;
 public abstract class BindingDescriptor<C extends StandardCredentials> extends Descriptor<MultiBinding<C>> {
 
     protected abstract Class<C> type();
+
+    /**
+     * Returns the context {@link MultiBinding} needs to access. Defaults to {@link Run}, {@link FilePath},
+     * {@link Launcher} and {@link TaskListener}.
+     *
+     * This allows the system to statically infer which steps are applicable in which context
+     * (say in freestyle or in workflow).
+     * @see StepContext#get(Class)
+     */
+    public Set<? extends Class<?>> getRequiredContext() {
+        return Collections.unmodifiableSet(new HashSet<Class<?>>(Arrays.asList(Run.class, FilePath.class, Launcher.class,
+                TaskListener.class)));
+    }
 
     public ListBoxModel doFillCredentialsIdItems(@AncestorInPath Item owner) {
         if (owner == null || !owner.hasPermission(Item.CONFIGURE)) {
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/BindingDescriptor.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/BindingDescriptor.java
@@ -29,21 +29,13 @@ import com.cloudbees.plugins.credentials.CredentialsProvider;
 import com.cloudbees.plugins.credentials.common.AbstractIdCredentialsListBoxModel;
 import com.cloudbees.plugins.credentials.common.StandardCredentials;
 import com.cloudbees.plugins.credentials.domains.DomainRequirement;
-import hudson.FilePath;
-import hudson.Launcher;
 import hudson.model.Descriptor;
 import hudson.model.Item;
-import hudson.model.Run;
-import hudson.model.TaskListener;
 import hudson.security.ACL;
 import hudson.util.ListBoxModel;
 
-import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashSet;
-import java.util.Set;
 
-import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.kohsuke.stapler.AncestorInPath;
 
 /**
@@ -55,16 +47,10 @@ public abstract class BindingDescriptor<C extends StandardCredentials> extends D
     protected abstract Class<C> type();
 
     /**
-     * Returns the context {@link MultiBinding} needs to access. Defaults to {@link Run}, {@link FilePath},
-     * {@link Launcher} and {@link TaskListener}.
-     *
-     * This allows the system to statically infer which steps are applicable in which context
-     * (say in freestyle or in workflow).
-     * @see StepContext#get(Class)
+     * Determines whether this {@link MultiBinding} needs a workspace to evaluate.
      */
-    public Set<? extends Class<?>> getRequiredContext() {
-        return Collections.unmodifiableSet(new HashSet<Class<?>>(Arrays.asList(Run.class, FilePath.class, Launcher.class,
-                TaskListener.class)));
+    public boolean requiresWorkspace() {
+        return true;
     }
 
     public ListBoxModel doFillCredentialsIdItems(@AncestorInPath Item owner) {
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
@@ -24,7 +24,6 @@
 
 package org.jenkinsci.plugins.credentialsbinding.impl;
 
-import com.google.inject.Inject;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.FilePath;
@@ -40,30 +39,36 @@ import java.io.ObjectStreamException;
 import java.io.OutputStream;
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import org.apache.commons.codec.Charsets;
 import org.jenkinsci.plugins.credentialsbinding.MultiBinding;
-import org.jenkinsci.plugins.workflow.steps.AbstractStepDescriptorImpl;
-import org.jenkinsci.plugins.workflow.steps.AbstractStepExecutionImpl;
-import org.jenkinsci.plugins.workflow.steps.AbstractStepImpl;
 import org.jenkinsci.plugins.workflow.steps.BodyExecutionCallback;
 import org.jenkinsci.plugins.workflow.steps.BodyInvoker;
 import org.jenkinsci.plugins.workflow.steps.EnvironmentExpander;
+import org.jenkinsci.plugins.workflow.steps.MissingContextVariableException;
+import org.jenkinsci.plugins.workflow.steps.Step;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
-import org.jenkinsci.plugins.workflow.steps.StepContextParameter;
+import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
+import org.jenkinsci.plugins.workflow.steps.StepExecution;
 import org.kohsuke.stapler.DataBoundConstructor;
+
+import javax.annotation.Nonnull;
 
 /**
  * Workflow step to bind credentials.
  */
 @SuppressWarnings("rawtypes") // TODO DescribableHelper does not yet seem to handle List<? extends MultiBinding<?>> or even List<MultiBinding<?>>
-public final class BindingStep extends AbstractStepImpl {
+public final class BindingStep extends Step {
 
     private final List<MultiBinding> bindings;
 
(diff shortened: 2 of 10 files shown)


src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -75,20 +80,41 @@
...
 
         @Override public boolean start() throws Exception {
+            Run<?,?> run = getContext().get(Run.class);
+            if (run == null) {
+                throw new MissingContextVariableException(Run.class);
+            }
+
+            FilePath workspace = getContext().get(FilePath.class);
+            Launcher launcher = getContext().get(Launcher.class);
+            TaskListener listener = getContext().get(TaskListener.class);
```
rsandell: why not fail quickly here as well since you claim to require a `TaskListener` below in `getRequiredContext`?

abayer: Good point, fixing.

src/main/java/org/jenkinsci/plugins/credentialsbinding/BindingDescriptor.java
```diff
@@ -44,6 +54,19 @@
 
     protected abstract Class<C> type();
 
+    /**
+     * Returns the context {@link MultiBinding} needs to access. Defaults to {@link Run}, {@link FilePath},
+     * {@link Launcher} and {@link TaskListener}.
```
jglick: Oxford comma!!!

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -75,20 +80,44 @@
         return bindings;
     }
 
-    public static final class Execution extends AbstractStepExecutionImpl {
+    @Override
+    public StepExecution start(StepContext context) throws Exception {
+        return new Execution(this, context);
+    }
+
+    public static final class Execution extends StepExecution {
```
jglick: Need no longer be `public`.

jglick: For running-build compatibility you must continue to extend `AbstractStepExecutionImpl`; see the Javadoc for nondeprecated constructor.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -75,20 +80,44 @@
...
+        private transient BindingStep step;
+
+        public Execution(@Nonnull BindingStep step, StepContext context) {
+            super(context);
+            this.step = step;
+        }
 
         @Override public boolean start() throws Exception {
+            Run<?,?> run = getContext().get(Run.class);
+            if (run == null) {
```
jglick: Impossible—we have declared this in `getRequiredContext`.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -75,20 +80,44 @@
...
+                throw new MissingContextVariableException(Run.class);
+            }
+            TaskListener listener = getContext().get(TaskListener.class);
+            if (listener == null) {
```
jglick: Ditto.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -177,9 +206,14 @@ private Object readResolve() throws ObjectStreamException {
 
         @Override protected void finished(StepContext context) throws Exception {
             Exception xx = null;
+            Run<?,?> run = context.get(Run.class);
+            if (run == null) {
```
jglick: ditto

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -75,20 +80,44 @@
...
             Map<String,String> overrides = new HashMap<String,String>();
             List<MultiBinding.Unbinder> unbinders = new ArrayList<MultiBinding.Unbinder>();
             for (MultiBinding<?> binding : step.bindings) {
+                for (Class<?> requiredContext : binding.getDescriptor().getRequiredContext()) {
+                    Object v = getContext().get(requiredContext);
+                    if (v == null) {
+                        throw new MissingContextVariableException(requiredContext);
+                    }
+                }
                 MultiBinding.MultiEnvironment environment = binding.bind(run, workspace, launcher, listener);
```
jglick: Hmm. So `MultiBinding.bind`’s signature should be changed to make `workspace` and `launcher` `@Nullable`, with Javadoc explaining whether the implementation can or cannot rely on them being present. Same for `Unbinder`. (`listener` should be marked `@Nonnull` like `build` already was.)

src/main/java/org/jenkinsci/plugins/credentialsbinding/BindingDescriptor.java
```diff
@@ -44,6 +54,19 @@
...
+     *
+     * This allows the system to statically infer which steps are applicable in which context
+     * (say in freestyle or in workflow).
+     * @see StepContext#get(Class)
+     */
+    public Set<? extends Class<?>> getRequiredContext() {
```
jglick: This does not really make sense as an API. There are two legitimate choices:

* `Run` + `TaskListener` + `FilePath` + `Launcher` (the default)
* `Run` + `TaskListener` (the new mode)

You could not usefully drop, say, `Run` or `TaskListener`—they will still be required. And you could not meaningfully request some additional context—you will indeed make the step fail if not called in that context, but you have no way of accessing the contextual object! (Since a `MultiBinding.bind` just gets fixed parameters.)

So I would just scrap this API and replace it with a simple

```java
public boolean requiresWorkspace() {
    return true;
}
```

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
```diff
@@ -165,6 +165,75 @@
...
+        final String username = "bob";
+        final String password = "s3cr3t";
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                UsernamePasswordCredentialsImpl c = new UsernamePasswordCredentialsImpl(CredentialsScope.GLOBAL, credentialsId, "sample", username, password);
+                CredentialsProvider.lookupStores(story.j.jenkins).iterator().next().addCredentials(Domain.global(), c);
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                p.setDefinition(new CpsFlowDefinition(""
+                        + "withCredentials([usernamePassword(usernameVariable: 'USERNAME', passwordVariable: 'PASSWORD', credentialsId: '" + credentialsId + "')]) {\n"
+                        + "  node {\n"  // We still need to enter a node to get the actual creds via the file.
```
jglick: No need. Try something like

```groovy
withCredentials([usernameColonPassword(variable: 'USERPASS', credentialsId: '…')]) {
  semaphore '…'
  echo THEVAR.toUpperCase()
}
```

which should print `BOB:S3CR3T`.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
```diff
@@ -165,6 +165,75 @@
...
+                CredentialsProvider.lookupStores(story.j.jenkins).iterator().next().addCredentials(Domain.global(), c);
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "p");
+                p.setDefinition(new CpsFlowDefinition(""
+                        + "withCredentials([file(variable: 'targetFile', credentialsId: '" + credentialsId + "')]) {\n"
+                        + "  echo 'We should fail before getting here'\n"
+                        + "}", true));
+                WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+                story.j.assertBuildStatus(Result.FAILURE, story.j.waitForCompletion(b));
+                story.j.assertLogNotContains("We should fail before getting here", b);
+                // We can't guarantee whether this will fail due to missing FilePath.class or Launcher.class.
```
jglick: Well if you adopt my suggested API you _can_ predict that.

src/main/java/org/jenkinsci/plugins/credentialsbinding/BindingDescriptor.java
abayer: =)

src/main/java/org/jenkinsci/plugins/credentialsbinding/BindingDescriptor.java
abayer: Works for me.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
abayer: Okiedokie.

abayer: Owie. Okiedokie.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
abayer: Yup, cleaning up.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
abayer: Gotcha.

abayer: Wait, `@Nullable`, not `@CheckForNull`?

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
abayer: Ah-ha - nice.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
abayer: =)

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
jglick: Yes—whether or not an implementer actually needs to check for null depends on factors outside of the reasonable scope of static analysis: how they implemented another method.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -85,38 +86,29 @@ public StepExecution start(StepContext context) throws Exception {
...
             Map<String,String> overrides = new HashMap<String,String>();
             List<MultiBinding.Unbinder> unbinders = new ArrayList<MultiBinding.Unbinder>();
             for (MultiBinding<?> binding : step.bindings) {
-                for (Class<?> requiredContext : binding.getDescriptor().getRequiredContext()) {
-                    Object v = getContext().get(requiredContext);
-                    if (v == null) {
-                        throw new MissingContextVariableException(requiredContext);
-                    }
+                if (binding.getDescriptor().requiresWorkspace() && workspace == null) {
+                    throw new MissingContextVariableException(FilePath.class);
```
jglick: Technically you should also check for `Launcher`, though in practice they will both be present or neither.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/FileBinding.java
```diff
@@ -81,14 +89,22 @@
...
+                                     @Nonnull TaskListener listener) throws IOException, InterruptedException {
             secretsDir(workspace).child(dirName).deleteRecursive();
         }
         
     }
 
-    private static FilePath secretsDir(FilePath workspace) {
-        return tempDir(workspace).child("secretFiles");
+    @CheckForNull
+    private static FilePath secretsDir(@CheckForNull FilePath workspace) {
```
jglick: Revert. It will not be null.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
abayer: Fair point!

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/FileBinding.java
```diff
@@ -51,24 +53,30 @@
...
-        dir.mkdirs();
-        secrets.chmod(/*0700*/448);
-        FilePath secret = dir.child(credentials.getFileName());
-        copy(secret, credentials);
-        if (secret.isDirectory()) { /* ZipFileBinding */
-            // needs to be writable so we can delete its contents
-            // needs to be executable so we can list the contents
-            secret.chmod(0700);
+        if (secrets == null) {
+            throw new IOException("Can't proceed with null workspace");
```
jglick: Revert. It cannot be null.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/FileBinding.java
abayer: Ok.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/FileBinding.java
abayer: Ok.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
abayer: Ok.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -107,7 +107,8 @@ public StepExecution start(StepContext context) throws Exception {
             Map<String,String> overrides = new HashMap<String,String>();
             List<MultiBinding.Unbinder> unbinders = new ArrayList<MultiBinding.Unbinder>();
             for (MultiBinding<?> binding : step.bindings) {
-                if (binding.getDescriptor().requiresWorkspace() && workspace == null) {
+                if (binding.getDescriptor().requiresWorkspace() &&
+                        (workspace == null || launcher == null)) {
                     throw new MissingContextVariableException(FilePath.class);
```
jglick: Misleading in case `FilePath` but not `Launcher` is present; again this is only theoretical.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -177,6 +199,7 @@ private Object readResolve() throws ObjectStreamException {
 
         @Override protected void finished(StepContext context) throws Exception {
             Exception xx = null;
+
```
jglick: gratuitous

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/FileBinding.java
```diff
@@ -62,8 +69,7 @@
             // needs to be writable so we can delete its contents
             // needs to be executable so we can list the contents
             secret.chmod(0700);
-        }
-        else {
+        } else {
```
jglick: gratuitous

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
abayer: Yeah - I decided to go with `FilePath` as the one I errored on fairly arbitrarily.

PR #35, opened 2017-04-27
[JENKINS-41760] Broken output when empty secret

mslusarczyk:
Fix for [JIRA](https://issues.jenkins-ci.org/browse/JENKINS-41760)


jglick:
cf. #43

PR #37, opened 2017-06-02
Close logger to avoid file descriptor leaks

mssbg:
A fix for https://issues.jenkins-ci.org/browse/JENKINS-43199


oleg-nenashev:
@jglick The issue is also with the bad design of logger decorators. And their default behavior... 
But I think it is still acceptable as a hotfix for older core versions

jglick:
Until there is a core PR fixing this properly I would not consider hot fixes.

kylebevans:
Hi everyone,

We're still struggling with this issue, and our master has to be restarted every few days.  What exactly is needed for a core PR?

oleg-nenashev:
I will let @jglick to respond, because I disagree with his response. He is the plugin maintainer, so he makes the final decision.

oleg-nenashev:
Just to clarify, I am blocking the PR only due to the imports, not due to the content.
@jglick It would be great to get some progress here. Even if we can create new logic in the core, a hotfix for older versions is definitely required. And IMHO this fix is really valid

mssbg:
I've restored the import order.

PR #41, opened 2017-06-21
POM updates

jglick:
Using a new parent POM, and updating some test dependencies.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.17</version>
+    <version>2.30</version>
     <relativePath />
   </parent>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.17</version>
+    <version>2.30</version>
```
jglick: This is the important part: it picks up https://github.com/jenkinsci/jenkins-test-harness/pull/35 and thus avoids the error described in https://github.com/jenkinsci/lib-annotation-indexer/pull/4 when running in `plugin-compat-tester`.

jglick:
> Could go higher on structs

Just picking the minimum needed to satisfy other dependencies.

PR #44, opened 2017-08-12
[JENKINS-44860] Username in a UsernamePasswordMultiBinding should not be considered sensitive

Stovoy:
We have credentials such as slack usernames and passwords. The usernames would match commonly occurring strings in the console output, and there'd be '****' everywhere, even though it's not sensitive information.

This attempts to fix that. If there is an argument that usernames are sensitive in some cases, perhaps this should go a bit further and add a checkbox to mark that the username is sensitive.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
@@ -80,19 +80,35 @@ public abstract class MultiBinding<C extends StandardCredentials> extends Abstra
     public static final class MultiEnvironment implements Serializable {
 
         private final Map<String,String> values;
+        private final Map<String,String> secretValues;
         private final Unbinder unbinder;
 
-        public MultiEnvironment(Map<String,String> values) {
-            this(values, new NullUnbinder());
+        public MultiEnvironment(Map<String,String> secretValues) {
+            this(secretValues, Collections.<String, String>emptyMap());
         }
 
-        public MultiEnvironment(Map<String,String> values, Unbinder unbinder) {
-            this.values = new HashMap<String,String>(values);
+        public MultiEnvironment(Map<String,String> secretValues, Map<String,String> publicValues) {
+            this(secretValues, publicValues, new NullUnbinder());
+        }
+
+        public MultiEnvironment(Map<String,String> secretValues, Unbinder unbinder) {
+            this(secretValues, Collections.<String, String>emptyMap(), unbinder);
+        }
+
+        public MultiEnvironment(Map<String,String> secretValues, Map<String,String> publicValues, Unbinder unbinder) {
+            this.secretValues = new HashMap<String,String>(secretValues);
+            this.values = new HashMap<String,String>();
+            this.values.putAll(secretValues);
+            this.values.putAll(publicValues);
             this.unbinder = unbinder;
         }
 
         public Map<String,String> getValues() {
             return Collections.unmodifiableMap(values);
+        }
+
+        public Map<String,String> getSecretValues() {
+            return Collections.unmodifiableMap(secretValues);
         }
 
         public Unbinder getUnbinder() {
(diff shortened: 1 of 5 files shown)


jglick:
This is [JENKINS-44860](https://issues.jenkins-ci.org/browse/JENKINS-44860) I think.

src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
```diff
@@ -80,21 +80,37 @@ public final String getCredentialsId() {
     public static final class MultiEnvironment implements Serializable {
 
         private final Map<String,String> values;
+        private final Map<String,String> secretValues;
```
jglick: This is going to be null after deserialization following a restart with a running Pipeline build, leading to `NullPointerException`s.

src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
```diff
@@ -80,21 +80,37 @@ public final String getCredentialsId() {
...
+        }
+
+        public MultiEnvironment(Map<String,String> secretValues, Unbinder unbinder) {
+            this(secretValues, Collections.<String, String>emptyMap(), unbinder);
+        }
+
+        public MultiEnvironment(Map<String,String> secretValues, Map<String,String> publicValues, Unbinder unbinder) {
+            this.secretValues = new HashMap<String,String>(secretValues);
+            this.values = new HashMap<String,String>();
+            this.values.putAll(secretValues);
```
jglick: So we store two copies of the secret values? That does not seem right.

I would suggest deprecating the `values` field and introducing a `publicValues` field; the constructor would initialize `secretValues` and `publicValues`, with `readResolve` doing the migration for the corner case.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/UsernamePasswordMultiBinding.java
```diff
@@ -74,14 +74,15 @@ public String getPasswordVariable() {
...
         StandardUsernamePasswordCredentials credentials = getCredentials(build);
-        Map<String,String> m = new HashMap<String,String>();
-        m.put(usernameVariable, credentials.getUsername());
-        m.put(passwordVariable, credentials.getPassword().getPlainText());
-        return new MultiEnvironment(m);
+        Map<String,String> secretValues = new HashMap<String,String>();
+        Map<String,String> publicValues = new HashMap<String,String>();
+        secretValues.put(passwordVariable, credentials.getPassword().getPlainText());
+        publicValues.put(usernameVariable, credentials.getUsername());
+        return new MultiEnvironment(secretValues, publicValues);
```
jglick: More simply:

```java
return new MultiEnvironment(Collections.singletonMap(passwordVariable, credentials.getPassword().getPlainText()), Collections.singletonMap(usernameVariable, credentials.getUsername()));
```

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/UsernamePasswordMultiBinding.java
```diff
@@ -74,14 +74,15 @@ public String getPasswordVariable() {
...
     }
 
     @Override public Set<String> variables() {
-        return new HashSet<String>(Arrays.asList(usernameVariable, passwordVariable));
+        return new HashSet<String>(Arrays.asList(passwordVariable));
```
jglick: I suppose this fixes `getSensitiveBuildVariables`, but then Javadoc for `variables()` should be updated to reflect that.

src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/UsernamePasswordMultiBindingTest.java
```diff
@@ -77,7 +77,7 @@
         FreeStyleBuild b = r.buildAndAssertSuccess(p);
         r.assertLogNotContains(password, b);
```
jglick: Could now add:

```java
r.assertLogContains(username, b);
```

jglick:
see #131

PR #47, opened 2018-02-06
Avoid NPE when usernamevar not specified

nwillems:
When using Jenkinsfiles, you can use the multibinding while avoiding to
specify either username or password. Seeing that you probably would
wan't the password in all cases, special treatment of the username has
been added.

This commit doesn't add testing, since I can't seem to reproduce the
problem in the testing instance.

Any recommendations on this are very welcome :-)


nwillems:
Hey @jglick are you still active on this plugin?

jglick:
Superseded by #54.

jglick:
Superseded by #54.

PR #50, opened 2018-07-14
[JENKINS-44860] Disable masking of usernames

henrist:
Carry of #44 

Fixes [JENKINS-44860](https://issues.jenkins-ci.org/browse/JENKINS-44860)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java
@@ -79,8 +79,10 @@ public abstract class MultiBinding<C extends StandardCredentials> extends Abstra
     /** Result of {@link #bind}. */
     public static final class MultiEnvironment implements Serializable {
 
-        private final Map<String,String> values;
-        private final Map<String,String> secretValues;
+        @Deprecated
+        private transient Map<String,String> values;
+        private Map<String,String> secretValues;
+        private Map<String,String> publicValues;
         private final Unbinder unbinder;
 
         public MultiEnvironment(Map<String,String> secretValues) {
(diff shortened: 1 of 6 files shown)


henrist:
@jglick Most of it is based on your previous review for #44 from January 10th, and the changes you requested there.

If I read your suggestion correctly, you propose the username should not be provided at all, which i can't see solves the issue in question. That would require hardcoding of the username, instead of using it from the credentials store.

jglick:
> you propose the username should not be provided at all, which […] would require hardcoding of the username

Yes. Seems close enough, especially given that there is a [trivial workaround](https://issues.jenkins-ci.org/browse/JENKINS-44860?focusedCommentId=353175&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-353175) for the case that you _do_ have to get the username from a credential store.

henrist:
I'm not sure we are on the same boat here @jglick.

You are talking about printing the username part from the credential on demand. Sure that can be worked around in different ways. I don't disagree on that.

My reason for investing time in this during last summer was to avoid build logs where "jenkins" is a natural part of the log, and at the same time credentials are using "jenkins" as usernames, leading to all occurrences of "jenkins" being masked when inside the `withCredentials` block.

If you have other ideas on how to resolve this, please share them, instead of pushing away people trying to contribute.

jglick:
> avoid build logs where "jenkins" is a natural part of the log, and at the same time credentials are using "jenkins" as usernames, leading to all occurrences of "jenkins" being masked when inside the `withCredentials` block

Yes this is certainly a problem, for which the simplest solution would be #54 (easier done than explained).

henrist:
And how does that solve the part that credentials often come as a pair of both username and password? I can't see how hardcoding the username helps.

Can you suggest how to resolve the original issue here? I think we keep talking around the issue finding up solutions for other problems.

jglick:
> I can't see how hardcoding the username helps.

Well, with #54, it solves the issue—again, only to the extent that the username is unlikely to change.

> Can you suggest how to resolve the original issue here?

#54 does resolve the “original issue” as stated

> it seems that not only are passwords masked, but usernames as well. This causes commonly used words (such as "jenkins", "android", "artifactory", "admin", etc.) to appear as **** in our console output, as we have credentials with usernames like that.

though not variants such as

> allow us to still provide the username from a Credential without it being obfuscated

which would still need to be worked around via (e.g.) `| od -a`.

henrist:
The issue states:
> Disable masking of usernames

You have provided the ability to not have usernames. Of cause, if you wanna play hard you can sure argue that solves the issue, but I'd call it a hack or a workaround more than a resolution for the issue in question.

But I don't understand what you want. You said the original PR had some minor issues but otherwise "looks right". You proposed changes. I followed up your proposals. And here you just ignore it telling you have _another_ suggestion. Great! I can't help any more. If you don't want to solve this then don't, I guess it's your call.

jglick:
I think we can do something like this PR as well, it is just a bigger code change that I would rather process myself. Sorry for the drive-by reviews; I have very little time to spend on any one of the many plugins for which I am owner by default, so lots of contributions languish unless they are very short and clearly harmless.

carlfriedrich:
@jglick @henrist What is the status here? The bug is really annoying, so I would like this to be merged.
Can I assist in any form?

henrist:
I can resolve the current conflicts if @jglick decides to move forward. But I think it's still uncertain what the maintainers actually want to do about this, so better agree on a plan before wasting more time.

carlfriedrich:
So @jglick, what is your plan here?

jglick:
I expect to revisit this + #44 + #54 at some point, taking into account effects in `pipeline-model-definition`.

jglick:
see #131

PR #53, opened 2018-12-03
Using GeneralNonBlockingStepExecution

jglick:
Downstream of https://github.com/jenkinsci/workflow-step-api-plugin/pull/38.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,7 +23,7 @@
     <jenkins.version>2.60.3</jenkins.version>
     <java.level>8</java.level>
     <workflow-cps-plugin.version>2.32</workflow-cps-plugin.version>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -5,23 +5,25 @@
...
   <name>Credentials Binding Plugin</name>
   <description>Allows credentials to be bound to environment variables for use from miscellaneous build steps.
   </description>
   <url>https://wiki.jenkins.io/display/JENKINS/Credentials+Binding+Plugin</url>
   <properties>
-    <jenkins.version>2.7.3</jenkins.version>
-    <java.level>7</java.level>
+    <revision>1.18</revision>
+    <changelist>-SNAPSHOT</changelist>
+    <jenkins.version>2.60.3</jenkins.version>
```
jvz: I know this is the baseline for Java 8, but do you think it's time to start relying on more recent LTS versions?

jglick: We could; there is no particular need to. Here I am just matching the baseline selected in the upstream PR, which _was_ to be able to use `-source 8`.

PR #56, opened 2019-01-18
Adapt tests to spaces in temp directories

jglick:
@thomasgl-orange alerted me in https://github.com/jenkinsci/jenkins-test-harness-htmlunit/pull/4#issuecomment-455370951 to the fact that https://github.com/jenkinsci/jenkins-test-harness/pull/89 breaks a test here.


jglick:
Hmm, broken CI.

PR #58, opened 2019-04-10
Java 11 readiness: build both on JDK8 and JDK11

batmat:
Making sure this plugin is constantly checked both building with JDK8 and JDK11.

This goes with the Jenkins Java 11 General Availability announcement made back in March, and we're making a pass to check plugins are looking good on a JDK11.

([to understand the Jenkinsfile content](https://wiki.jenkins.io/display/JENKINS/Java+11+Developer+Guidelines#Java11DeveloperGuidelines-MakesureyourpluginistestedinContinuousIntegrationonJava8andJava11atthesametime))

@jenkinsci/java11-support

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,11 +20,11 @@
   <properties>
     <revision>1.19</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.60.3</jenkins.version>
+    <jenkins.version>2.150.1</jenkins.version>
     <java.level>8</java.level>
     <workflow-cps-plugin.version>2.32</workflow-cps-plugin.version>
-    <workflow-step-api-plugin.version>2.18</workflow-step-api-plugin.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -148,6 +148,12 @@
       <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>script-security</artifactId>
```
jglick: Probably better put in `dependencyManagement`, since it is not a direct dep AFAIK, but OK.

pom.xml
```diff
@@ -157,7 +163,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>scm-api</artifactId>
```
jglick: Ditto this, but as you can see it was already here.

PR #60, opened 2019-05-30
Prototypes around better support for user-scoped credentials

jvz:
This work is to explore the possibilities around using and improving on user-scoped credentials in pipelines. They were originally developed and designed for free style projects, so this effort is to see how feasible it would be to extend those features to work better with pipelines.

As it stands now, it seems as though that this already sort of works when you use the authorize project plugin. However, it seems like that plugin has compatibility issues with other plugins, or people just can't figure out how to administer it correctly in practice.

Using user-scoped credentials via an input step or build parameter are alternatives to executing the pipeline as a specific user may work more compatibly. There is some support for this already in BindingStep before it looks up the credentials. I have not yet shown how that works in a test.

The final bit of feasibility to check here is credentials tracking. That may be useful for limiting which credentials are authorized to a running build based on user interactions (e.g., only allow access to credentials that were selected from credentials inputs).

There may also be some use in storing credentials to other ModelObjects besides User, Jenkins, and folders, though this may have its own credentials-plugin changes.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
@@ -24,38 +24,28 @@
 
 package org.jenkinsci.plugins.credentialsbinding.impl;
 
+import com.cloudbees.plugins.credentials.Credentials;
 import com.cloudbees.plugins.credentials.CredentialsProvider;
 import com.cloudbees.plugins.credentials.CredentialsScope;
 import com.cloudbees.plugins.credentials.SecretBytes;
 import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
 import com.cloudbees.plugins.credentials.domains.Domain;
 import com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl;
-
-import hudson.model.Fingerprint;
-import hudson.model.User;
-import jenkins.security.QueueItemAuthenticatorConfiguration;
-
 import hudson.FilePath;
 import hudson.Functions;
+import hudson.model.Fingerprint;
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
```diff
@@ -443,6 +440,80 @@ public void testTrackingOfCredential() {
...
+                User user = User.getOrCreateByIdOrFullName("beth");
+                String credentialsId = UUID.randomUUID().toString();
+                String secret = "correct horse battery staple";
+                registerUserCredentials(user, new StringCredentialsImpl(CredentialsScope.USER, credentialsId,
+                        "User-scoped credential data from Beth Unit", Secret.fromString(secret)));
+
+                WorkflowJob p = story.j.createProject(WorkflowJob.class);
+                p.setDefinition(new CpsFlowDefinition(
+                        "node {\n" +
+                                "  error 'TODO: use a build parameter to prompt for credentials and use the above credentials'\n" +
```
jglick: See [JENKINS-38963](https://issues.jenkins-ci.org/browse/JENKINS-38963) for a supposedly working syntax.

jvz: Yes, the dollar sign syntax thing is explicitly coded for in BindingStep. I plan to use that syntax in a manual test of parameterized builds, though I also hope to figure out how to create an automated test around it.

jvz:
Implemented in other prototypes via https://github.com/jenkinsci/credentials-plugin/pull/119

jvz:
Closing this PR as it's unnecessary to implement the improvements (which are mostly contained in credentials-plugin).

PR #63, opened 2019-08-01
Add support for Almquist shell in masking

Wadeck:
Almquist shell is the default one for Alpine. The escape schema is a bit special compared to the other.

```
pwd="pass'word"
set -x
echo $pwd
```
outputs
`+ echo 'pass'"'"'word'`

You can reproduce those commands on `jenkinsci/jnlp-slave:alpine` default shell.

Unit tests added.

For the review information: `'` is escaped as `'"'"'`, `''` as `'"''"'`, `'''` as `'"'''"'`

@jvz

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/masking/AlmquistShellSecretPatternFactory.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/masking/AlmquistShellSecretPatternFactory.java
@@ -1,0 +1,73 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2019 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jenkinsci.plugins.credentialsbinding.masking;
+
+import hudson.Extension;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+import javax.annotation.Nonnull;
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * Almquist shell, aka ash, is the default shell in Alpine distribution
+ */
+@Extension
+@Restricted(NoExternalUse.class)
+public class AlmquistShellSecretPatternFactory implements SecretPatternFactory {
+
+    private static final char SINGLE_QUOTE = '\'';
+    private static final String START_FRAGMENT = "'\"";
+    private static final String END_FRAGMENT = "\"'";
+
+    private @Nonnull String getQuotedForm(@Nonnull String input) {
+        StringBuilder sb = new StringBuilder(input.length());
+        for (int i = 0; i < input.length(); i++) {
+            char c = input.charAt(i);
+            if (c == SINGLE_QUOTE) {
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/credentialsbinding/masking/AlmquistShellSecretPatternFactoryTest.java
```diff
@@ -0,0 +1,140 @@
...
+        return passwords;
+    }
+
+    @ClassRule public static JenkinsRule j = new JenkinsRule();
+
+    private WorkflowJob project;
+    private String credentialsId;
+
+    @BeforeClass
+    public static void assumeBash() {
```
jglick:
```suggestion
    public static void assumeAsh() {
```

src/test/java/org/jenkinsci/plugins/credentialsbinding/masking/AlmquistShellSecretPatternFactoryTest.java
```diff
@@ -0,0 +1,140 @@
...
+        CredentialsTestUtil.setStringCredentials(j.jenkins, credentialsId, credentials);
+        WorkflowRun run = runProject();
+
+        j.assertLogContains("begin0 **** end0", run);
+        j.assertLogContains("begin2 **** end2", run);
+        j.assertLogNotContains(credentials, run);
+    }
+
+    private WorkflowRun runProject() throws Exception {
+        return j.assertBuildStatusSuccess(project.scheduleBuild2(0));
```
jglick:
```suggestion
        return j.buildAndAssertSuccess(project);
```

which is so short it should just be inlined anyway.

src/test/java/org/jenkinsci/plugins/credentialsbinding/masking/AlmquistShellSecretPatternFactoryTest.java
```diff
@@ -0,0 +1,140 @@
...
+    private WorkflowJob project;
+    private String credentialsId;
+
+    @BeforeClass
+    public static void assumeBash() {
+        // ash = Almquist shell, default one used in Alpine
+        assumeThat("ash", is(executable()));
+        // due to https://github.com/jenkinsci/durable-task-plugin/blob/e75123eda986f20a390d92cc892c3d206e60aefb/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java#L149
+        // on Windows
+        assumeThat("nohup", is(executable()));
```
jglick: Huh? Why is this here? The link does not explain anything to me.

src/main/java/org/jenkinsci/plugins/credentialsbinding/masking/AlmquistShellSecretPatternFactory.java
```diff
@@ -0,0 +1,73 @@
...
+import hudson.Extension;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+import javax.annotation.Nonnull;
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * Almquist shell, aka ash, is the default shell in Alpine distribution
```
jglick:
```suggestion
 * Almquist shell, aka ash, is the default shell in Alpine distribution.
```

(`javadoc` cares about punctuation.)

src/test/java/org/jenkinsci/plugins/credentialsbinding/masking/AlmquistShellSecretPatternFactoryTest.java
```diff
@@ -0,0 +1,140 @@
...
+import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.experimental.theories.DataPoint;
+import org.junit.experimental.theories.DataPoints;
+import org.junit.experimental.theories.Theories;
+import org.junit.experimental.theories.Theory;
```
jglick: Is there a particular reason you are using this “experimental” facility? AFAICT the test could be written just as well using the better-known `Parameters`.

Wadeck: Matt provided that approach in the previous test for Bash/Bourne shell, so I re-used them. I was a bit disappointed to see that the theories are failing as soon as one dataPoint failed compared to parameter.

This could be changed in a future version without problem (along with the other usage of that approach).

src/test/java/org/jenkinsci/plugins/credentialsbinding/masking/AlmquistShellSecretPatternFactoryTest.java
Wadeck: The link points to a code adding the `nohup` argument in the command. On Windows that command is not necessary present (at least not in my environments) and thus just checking for the shell is not sufficient when used in `durable-task`.

jglick: I would not expect `ash` to be present on Windows either.

Wadeck: Good point 😕

jglick: Anyway, it is OK, just a bit confusing.

jglick:
[INFRA-2075](https://issues.jenkins-ci.org/browse/INFRA-2075)

Wadeck:
Manually tested some different scenarios, you can merge it :)

PR #80, opened 2019-12-02
Do not mask secrets shorter than (3) characters.

jeffret-b:
Only mask secrets greater than a minimum number of characters. Masking really short secrets can play havoc with the log. For example, if the password value is "a", all instances of 'a' in the log will be replaced by the masking string. This makes the log unreadable.

The `MINIMUM_SECRET_MASKING_LENGTH` is arbitrary, but 3 seems like a good compromise to not mask too many things that shouldn't be while still masking meaningful secrets. It could be 2. Or 4. Making it user configurable is a bit excessive.

@jvz @Wadeck @daniel-beck

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/masking/SecretPatterns.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/masking/SecretPatterns.java
@@ -36,7 +36,7 @@ import java.util.stream.Collectors;
 @Restricted(NoExternalUse.class)
 public class SecretPatterns {
 
-    private static final int MINIMUM_PASSWORD_MASKING_LENGTH = 3;
+    private static final int MINIMUM_SECRET_MASKING_LENGTH = 3;
 
     private static final Comparator<String> BY_LENGTH_DESCENDING =
             Comparator.comparingInt(String::length).reversed().thenComparing(String::compareTo);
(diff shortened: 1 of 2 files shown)


jeffret-b:
This PR serves to highlight the fundamental question: Is adding a minimum length for masking a good idea? There are conflicting arguments with no one yet feeling confident in either side. Personally I lean towards not proceeding with this PR and leaving things as is.

We could add a configurable length as a property. That's largely a concession that we couldn't agree on an answer to the fundamental question or the next one of how long the minimum should be.

jeffret-b:
Without clear documentation that there's a need here, we should leave this as-is.

src/main/java/org/jenkinsci/plugins/credentialsbinding/masking/SecretPatterns.java
```diff
@@ -51,6 +53,7 @@
     public static @Nonnull Pattern getAggregateSecretPattern(@Nonnull Collection<String> inputs) {
         String pattern = inputs.stream()
                 .filter(input -> !input.isEmpty())
+                .filter(input -> input.length() > MINIMUM_SECRET_MASKING_LENGTH)
```
jvz: This filter can replace the one above since they're both string length checks.

jeffret-b:
After lengthy discussions in our team, we decided to close this, at least for now. We haven't identified a clear need. The preference of whether to add this or not along with the minimum number are arbitrary. We could add a configurable parameter for the count, but that highlights rather than reducing the arbitrariness.

This could be re-opened if someone provides better justification or clarity.

PR #88, opened 2020-02-24
Demonstrate adding findsecbugs to spotbugs.

jeffret-b:
Demonstrate adding findsecbugs to spotbugs. This commit is a demonstration and is not intended to be merged as-is.

Similar to jenkinsci/platformlabeler-plugin#165, findsecbugs finds no issues in this plugin. Nothing will need to be done here when the parent pom is updated to include findsecbugs.

---

My plan is to add findsecbugs at the plugin pom after sufficient testing and communication.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -145,4 +145,22 @@
     </dependency>
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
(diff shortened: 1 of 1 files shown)


jglick:
Sounds good.

PR #92, opened 2020-04-03
Add some documentation about not masking multi-line secrets.

jeffret-b:
There has been some uncertainty or confusion about whether multi-line secrets are expected to be masked. These types of secrets include things like SSH private key and various types of certificates. This PR adds a sentence in a few places to clarify that multi-line secrets are not masked.


jeffret-b:
@jvz @daniel-beck @Wadeck @jglick , please take a look.

PR #105, opened 2020-07-28
[JENKINS-63254][JENKINS-47101] Watch sensitive variables

car-roll:
[JENKINS-63254](https://issues.jenkins-ci.org/browse/JENKINS-63254)

PoC version 2 for groovy string interpolation interception.
Relies on changes to `EnvironmentExpander`. Unlike version 1, does not require any changes to core.

Requires: https://github.com/jenkinsci/workflow-step-api-plugin/pull/57
Upstream: https://github.com/jenkinsci/workflow-cps-plugin/pull/370

TODO:
~- Add unit test(s)~

CC @dwnusbaum @bitwiseman

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
     <changelist>-SNAPSHOT</changelist>
     <jenkins.version>2.176.4</jenkins.version>
     <java.level>8</java.level>
-    <workflow-step-api.version>2.23-rc567.0fe52fbdf6b5</workflow-step-api.version>
+    <workflow-step-api.version>2.23</workflow-step-api.version> <!-- TODO: Delete when in BOM -->
   </properties>
   <licenses>
     <license>
--- a/pom.xml
+++ b/pom.xml
@@ -73,7 +73,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-step-api</artifactId>
-      <version>2.23-rc565.4e7dd9b85962</version>
+      <version>2.23-rc566.c1fa948b49be</version>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
@@ -112,7 +112,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-step-api</artifactId>
-      <version>2.23-rc565.4e7dd9b85962</version>
+      <version>2.23-rc566.c1fa948b49be</version>
       <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -22,6 +22,7 @@
     <changelist>-SNAPSHOT</changelist>
     <jenkins.version>2.176.4</jenkins.version>
     <java.level>8</java.level>
+    <workflow-step-api.version>2.23-rc567.0fe52fbdf6b5</workflow-step-api.version>
   </properties>
   <licenses>
     <license>
@@ -73,7 +74,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-step-api</artifactId>
-      <version>2.23-rc566.c1fa948b49be</version>
+      <version>${workflow-step-api.version}</version>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -181,10 +182,15 @@ private void doStart() throws Exception {
...
             for (Map.Entry<String,Secret> override : overrides.entrySet()) {
-                env.override(override.getKey(), override.getValue().getPlainText());
+                String keyOverride = override.getKey();
+                env.override(keyOverride, override.getValue().getPlainText());
+                sensitiveVars.add(keyOverride);
             }
         }
 
+        @Override public Set<String> getSensitiveVars() {
+            return Collections.unmodifiableSet(sensitiveVars);
```
dwnusbaum, line 192: I think you could simplify things so you don't even have to add the `sensitiveVars` field:

```suggestion
            return Collections.unmodifiableSet(overrides.keySet());
```

I think it is normal or at least possible for `expand` to be called more than once on different instances of `EnvVars`, and the current code would try to update `sensitiveVars` every time `expand` is called, which would have no effect since the field is a `Set`, but is a bit redundant. 

Also, the current implementation means that `getSensitiveVars` returns an empty set until `expand` has been called at least once, which in practice might be fine but it seems like it could cause confusion at some point; using `overrides.getKeySet()` avoids this problem.

car-roll: regarding the deserialization problem, would it be enough to add `null`  checks whenever `sensitiveVars` is accessed?

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -168,6 +168,7 @@ private void doStart() throws Exception {
     }
 
     private static final class Overrider extends EnvironmentExpander {
+        private Set<String> sensitiveVariables = new HashSet<>();
```
dwnusbaum, line 171:
```suggestion
```

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -181,10 +182,15 @@ private void doStart() throws Exception {
 
         @Override public void expand(EnvVars env) throws IOException, InterruptedException {
             for (Map.Entry<String,Secret> override : overrides.entrySet()) {
-                env.override(override.getKey(), override.getValue().getPlainText());
+                String keyOverride = override.getKey();
+                env.override(keyOverride, override.getValue().getPlainText());
+                sensitiveVariables.add(keyOverride);
```
dwnusbaum, line 187:
```suggestion
                env.override(override.getKey(), override.getValue().getPlainText());
```

car-roll: addressed in 06428a0516322ded27dffb274fdcf04017ab944e

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
car-roll: removed in 06428a0516322ded27dffb274fdcf04017ab944e

dwnusbaum (approved):

bitwiseman (approved):

pom.xml
```diff
@@ -20,8 +20,9 @@
   <properties>
     <revision>1.24</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.164.3</jenkins.version>
+    <jenkins.version>2.176.4</jenkins.version>
     <java.level>8</java.level>
+    <workflow-step-api.version>2.23-rc567.0fe52fbdf6b5</workflow-step-api.version>
```
jglick, line 25:
```suggestion
    <workflow-step-api.version>2.23-rc567.0fe52fbdf6b5</workflow-step-api.version> <!-- TODO https://github.com/jenkinsci/workflow-step-api-plugin/pull/57; delete when in BOM -->
```

jglick: Or once released:
```suggestion
    <workflow-step-api.version>2.23</workflow-step-api.version> <!-- TODO delete when in BOM -->
```

pom.xml
```diff
@@ -20,8 +20,9 @@
   <properties>
     <revision>1.24</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.164.3</jenkins.version>
+    <jenkins.version>2.176.4</jenkins.version>
```
jglick, line 23: You forgot to update the `bom` to match.

jglick (changes requested):

jglick (approved):

PR #107, opened 2020-08-10
[JENKINS-63366] Apply credentials masking first

jvz:
This rearranges the ordering of the configured log filters for a step by
applying the newly created filter as the first log filter instead of
last.

See [JENKINS-63366](https://issues.jenkins-ci.org/browse/JENKINS-63366).

@jglick @Wadeck @daniel-beck @jeffret-b

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
@@ -35,12 +35,13 @@ import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.util.Secret;
 import java.io.IOException;
+import java.io.InvalidObjectException;
 import java.io.ObjectStreamException;
 import java.io.OutputStream;
 import java.io.Serializable;
+import java.net.UnknownServiceException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -187,53 +193,95 @@ private void doStart() throws Exception {
...
+    }
+
+    /** @deprecated Use {@link MaskingDecorator} */
+    @Deprecated
+    private static final class Filter extends ConsoleLogFilter implements Serializable {
+
+        private static final long serialVersionUID = 1;
+
+        @Override public OutputStream decorateLogger(AbstractBuild _ignore, final OutputStream logger) throws IOException, InterruptedException {
+            throw new UnknownServiceException();
```
jglick, line 283: ??

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -187,53 +193,95 @@ private void doStart() throws Exception {
...
+            return new SerializationProxy(Secret.fromString(pattern.pattern()));
+        }
+
         private Object readResolve() throws ObjectStreamException {
-            if (this.charsetName == null) {
-                this.charsetName = Charsets.UTF_8.name();
+            throw new InvalidObjectException("Must use proxy");
+        }
+
+        private static class SerializationProxy implements Serializable {
```
jglick, line 217: Seems like overkill. Just have `SecretPattern` compile its pattern on demand.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
jvz: Typo most likely; I must've hit tab after typing `throw new Un`. Should be `UnsupportedOperationException`.

jglick: I think you must retain compatibility here. If someone has a build running with credential masking active, upgrades the plugin, and restarts Jenkins, this method will be called after the restart.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
jglick: You just have

```java
private synchronized Pattern pattern() {
  if (pattern == null) {
    pattern = Pattern.compile(secret);
  }
  return pattern;
}
```

so it is compiled at most once per actual instance of `SecretPattern`.

jvz: Then I think this class is only here due to best practices with serialization (i.e., use a serialization proxy unless your class is a pure data POJO of some sort with no enforced invariants).

jglick: Just suggesting a simplification that avoids the need for `writeReplace` / `readResolve` tricks.

PR #110, opened 2020-09-28
Highlight importance of using single-quotes.

jeffret-b:
Increase the documented visibility of the recommendation to use single quotes whenever possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,11 +45,11 @@ withCredentials([usernamePassword(credentialsId: 'amazon', usernameVariable: 'US
 
 #### Note
 
-You should use a single quote (') instead of a double quote (") whenever you can. 
+You should use a single quote (`'`) instead of a double quote (`"`) whenever you can. 
 This is particularly important in Pipelines where a statement may be interpreted by both the Pipeline engine and an external interpreter, such as a Unix shell (`sh`) or Windows Command (`bat`) or Powershell (`ps`). 
 This reduces complications with password masking and command processing. 
 The first step in the above example properly demonstrates this. 
-The next two steps use the basic Pipeline "echo" step. 
+The next two steps use the basic Pipeline `echo` step. 
 The first one references the Groovy variable and needs no quotes. 
 The second one needs to use double quotes, so that the interpolation is performed in Groovy.
 
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ withCredentials([usernamePassword(credentialsId: 'amazon', usernameVariable: 'US
 #### Note
 
 You should use a single quote (') instead of a double quote (") whenever you can. 
-This is particularly important in Pipelines where a statement may be interpreted by both the Pipeline engine and an external interpreter, such as a Unix shell or Windows Command or Powershell. 
+This is particularly important in Pipelines where a statement may be interpreted by both the Pipeline engine and an external interpreter, such as a Unix shell (`sh`) or Windows Command (`bat`) or Powershell (`ps`). 
 This reduces complications with password masking and command processing. 
 The first step in the above example properly demonstrates this. 
 The next two steps use the basic Pipeline "echo" step. 
--- a/README.md
+++ b/README.md
@@ -43,6 +43,18 @@ withCredentials([usernamePassword(credentialsId: 'amazon', usernameVariable: 'US
 }
 ```
 
+#### Note
+
+You should use a single quote (') instead of a double quote (") whenever you can. 
+This is particularly important in Pipelines where a statement may be interpreted by both the Pipeline engine and an external interpreter, such as a Unix shell or Windows Command or Powershell. 
+This reduces complications with password masking and command processing. 
+The first step in the above example properly demonstrates this. 
+The next two steps use the basic Pipeline "echo" step. 
+The first one references the Groovy variable and needs no quotes. 
+The second one needs to use double quotes, so that the interpolation is performed in Groovy.
+
+For more information, see the Pipeline step reference for [Credentials Binding Plugin](https://www.jenkins.io/doc/pipeline/steps/credentials-binding/).
+
 ## Changelog
 
 See [GitHub Releases](https://github.com/jenkinsci/credentials-binding-plugin/releases) for new releases,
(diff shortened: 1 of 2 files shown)


jeffret-b:
@Wadeck, does this look like what you want?

README.md
```diff
@@ -43,6 +43,18 @@ withCredentials([usernamePassword(credentialsId: 'amazon', usernameVariable: 'US
 }
 ```
 
+#### Note
+
+You should use a single quote (') instead of a double quote (") whenever you can. 
+This is particularly important in Pipelines where a statement may be interpreted by both the Pipeline engine and an external interpreter, such as a Unix shell or Windows Command or Powershell.
```
Wadeck, line 49:
```suggestion
This is particularly important in Pipelines where a statement may be interpreted by both the Pipeline engine and an external interpreter, such as a Unix shell (`sh`) or Windows Command (`bat`) or Powershell (`ps`). 
```

src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep/help.html
```diff
@@ -120,3 +120,8 @@
     <a href="https://jenkins.io/blog/2019/02/21/credentials-masking/" target="_blank">Limitations of Credentials Masking</a>
     blog post for more background.
 </p>
+<h4>Note</h4>
+<p>
+    You should use a single quote (') instead of a double quote (") whenever you can. 
+    This reduces complications with password masking and command processing.
+</p>
```
Wadeck, line 127: No links to another article with more details

Wadeck (commented):
We have agreed to add a link to an article with details about password masking that could be broken depending on the content of the variables due to double quotes. This article is not yet written / up-to-date. 

So this PR without the expected links (in both readme and help) has low value as there is no explanation about why ("complications" is not enough in this particularly difficult topic).

Do you expect the article to be a new one or a rewamp of an existing one?

jeffret-b:
> Do you expect the article to be a new one or a revamp of an existing one?

I expect it to be a new one. There really isn't an existing one that it makes sense to revamp.

We could make that new document a blog post -- that format would make it easier to get something out more quickly, even if it's later followed up by more polished documentation.

README.md
```diff
@@ -43,6 +43,18 @@ withCredentials([usernamePassword(credentialsId: 'amazon', usernameVariable: 'US
 }
 ```
 
+#### Note
+
+You should use a single quote (') instead of a double quote (") whenever you can.
```
jvz, line 48:
```suggestion
You should use a single quote (') instead of a double quote (") whenever you can to provide data to pipeline steps. 
```

src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep/help.html
```diff
@@ -120,3 +120,8 @@
     <a href="https://jenkins.io/blog/2019/02/21/credentials-masking/" target="_blank">Limitations of Credentials Masking</a>
     blog post for more background.
 </p>
+<h4>Note</h4>
+<p>
+    You should use a single quote (') instead of a double quote (") whenever you can.
```
jvz, line 125:
```suggestion
    You should use a single quote (') instead of a double quote (") whenever you can for data provided to pipeline steps.
```

jvz (approved):

README.md
jeffret-b: I don't know. Isn't the current statement correct? -- You should use a single quote whenever you can. Does qualifying it like help the user understand better?

jvz: Oh I didn't think of it that way; good point. Seems fine the way you wrote it then.

Wadeck:
_Feedback provided in https://cloudbees.atlassian.net/browse/JENSEC-1069 (CloudBees internal ticket)_

README.md
jglick, line 48:
```suggestion
You should use a single quote (`'`) instead of a double quote (`"`) whenever you can. 
```

README.md
```diff
@@ -43,6 +43,18 @@ withCredentials([usernamePassword(credentialsId: 'amazon', usernameVariable: 'US
 }
 ```
 
+#### Note
+
+You should use a single quote (') instead of a double quote (") whenever you can. 
+This is particularly important in Pipelines where a statement may be interpreted by both the Pipeline engine and an external interpreter, such as a Unix shell (`sh`) or Windows Command (`bat`) or Powershell (`ps`). 
+This reduces complications with password masking and command processing. 
+The first step in the above example properly demonstrates this. 
+The next two steps use the basic Pipeline "echo" step.
```
jglick, line 52:
```suggestion
The next two steps use the basic Pipeline `echo` step. 
```

src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep/help.html
jglick, line 125:
```suggestion
    You should use a single quote (<code>'</code>) instead of a double quote (<code>"</code>) whenever you can. 
```

src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep/help.html
```diff
@@ -120,3 +120,8 @@
     <a href="https://jenkins.io/blog/2019/02/21/credentials-masking/" target="_blank">Limitations of Credentials Masking</a>
     blog post for more background.
 </p>
+<h4>Note</h4>
```
jglick, line 123: This note can just be deleted. It is already covered in https://github.com/jenkinsci/credentials-binding-plugin/blob/9abc42b55ae68d731749388aa4f2adbe4655b26b/src/main/resources/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep/help.html#L28-L42

jglick (commented):
https://github.com/jenkinsci/workflow-cps-plugin/pull/370 by @car-roll will probably help more.

jeffret-b:
> [jenkinsci/workflow-cps-plugin#370](https://github.com/jenkinsci/workflow-cps-plugin/pull/370) by @car-roll will probably help more.

Yes, I agree with that.

jeffret-b:
I updated this per Jesse's review comments, but it still needs coordination with other changes. It should be considered on hold for now.

jglick (approved):

PR #129, opened 2021-04-26
`maven-windows` agents seem broken

jglick:
Like https://github.com/jenkinsci/file-parameters-plugin/pull/44. CC @olblak, @MarkEWaite, @timja, not sure who manages this sort of thing?


timja:
can't see anything in the portal or ci.jenkins.io from a quick look, all seems to say it's fine =/

PR #141, opened 2021-07-01
[JENKINS-64361] Make fix for JENKINS-44860 apply to Pipeline step arguments as well

dwnusbaum:
See [JENKINS-64631](https://issues.jenkins.io/browse/JENKINS-64631).

#131 fixed [JENKINS-44860](https://issues.jenkins.io/browse/JENKINS-44860) by allowing usernames to remain unmasked in build logs, but it did not update the method introduced in #105 that Pipeline uses to know what environment variables contain secrets, so usernames were still being masked in Pipeline step arguments regardless of the value of `UsernameCredentials.isUsernameSecret`. This PR is a minor extension to #131 to address that issue.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #143, opened 2021-07-12
[JENKINS-64361] Make fix for JENKINS-44860 apply to Pipeline step arguments as well (take II)

jglick:
Should be the same effect as #141 but I think a little clearer to read. Also
* avoids wrapping a non-secret in `Secret`
* avoids storing secret variable names twice


src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
```diff
@@ -177,22 +172,34 @@ private void doStart() throws Exception {
         private static final long serialVersionUID = 1;
 
         private final Map<String,Secret> overrides = new HashMap<String,Secret>();
+        private Map<String, String> publicOverrides = new HashMap<>();
```
dwnusbaum, line 175: nit: The `new HashMap<>()` here is never used.

dwnusbaum (approved):

PR #146, opened 2021-07-23
Enable ssh keys to be written with user provided property

ricedavida:
The code change will take, if provided, the user property setting of an ssh key's file encoding, 
and it will set it on the remote agent's computer.  If left unset, it will use the default of UTF-8.

This feature is needed for edge systems that don't use UTF-8 natively, like zOS(the mainframe).  So, to allow the
ssh keyfiles to be generated correctly, we can specify the encoding type. 

This has been manually tested on a jenkins/zOS, and works as intended.


jglick:
Can this not just be set to ASCII?

PR #269, opened 2023-09-06
Add base64 masking

Kevin-CB:
There is no support for base64-encoded credentials, and it's not an encryption method and it can be easily decoded.

To prevent base64-encoded credentials from being exposed in build logs, I added a new SecretPatternFactory for it.

### Testing done

I wrote a test to ensure it was done properly on Windows and Unix-based systems but I also tested it locally with the following pipeline:
```
node {
    withCredentials([usernamePassword(credentialsId: "CredID1", usernameVariable: 'USERNAME', passwordVariable: 'PASSWORD')]) {
        sh '''
            echo Base64 encoded password:
            echo -n $PASSWORD | base64
            echo "========================================"
            curl -v https://$USERNAME:$PASSWORD@google.com
            echo "========================================"
        '''
    }
}
```
<details>
  <summary>Before:</summary>
  
Clear
</details>

<details>
  <summary>After:</summary>
  
Encrypted
</details>

```[tasklist]
### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/masking/Base64SecretPatternFactory.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/masking/Base64SecretPatternFactory.java
@@ -20,29 +20,28 @@ public class Base64SecretPatternFactory implements SecretPatternFactory {
     }
 
     @NonNull
-    private Collection<String> getBase64Forms(@NonNull String secret) {
+    public Collection<String> getBase64Forms(@NonNull String secret) {
         if (secret.length() == 0) {
             return Collections.emptyList();
         }
 
-        Base64.Encoder encoder = Base64.getEncoder();
-        Collection<String> result = new ArrayList<>(3);
+        Base64.Encoder[] encoders = new Base64.Encoder[]{
+                Base64.getEncoder(),
+                Base64.getUrlEncoder(),
+        };
 
-        // default
-        String regularBase64 = encoder.encodeToString(secret.getBytes(StandardCharsets.UTF_8));
-        result.add(regularBase64);
-        result.add(removeTrailingEquals(regularBase64));
+        Collection<String> result = new ArrayList<>();
+        String[] shifts = {"", "a", "aa"};
 
-        // shifted by one
-        String shiftedByOne = encoder.encodeToString(("a" + secret).getBytes(StandardCharsets.UTF_8));
-        result.add(shiftedByOne.substring(2));
-        result.add(removeTrailingEquals(shiftedByOne.substring(2)));
-
-        // shifted by two
-        String shiftedByTwo = encoder.encodeToString(("aa" + secret).getBytes(StandardCharsets.UTF_8));
-        result.add(shiftedByTwo.substring(4));
-        result.add(removeTrailingEquals(shiftedByTwo.substring(4)));
-
+        for (String shift : shifts) {
+            for (Base64.Encoder encoder : encoders) {
+                String shiftedSecret = shift + secret;
+                String encoded = encoder.encodeToString(shiftedSecret.getBytes(StandardCharsets.UTF_8));
+                String processedEncoded = shift.length() > 0 ? encoded.substring(2 * shift.length()) : encoded;
+                result.add(processedEncoded);
+                result.add(removeTrailingEquals(processedEncoded));
+            }
+        }
         return result;
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/credentialsbinding/masking/Base64SecretPatternFactory.java
```diff
@@ -0,0 +1,60 @@
...
+        return getBase64Forms(input);
+    }
+
+    @NonNull
+    private Collection<String> getBase64Forms(@NonNull String secret) {
+        if (secret.length() == 0) {
+            return Collections.emptyList();
+        }
+
+        Base64.Encoder encoder = Base64.getEncoder();
```
jglick, line 28: Maybe we should cover variants from `getUrlEncoder` as well? That would handle `+/` → `-_` which might be used on occasion.

(`getMimeEncoder` seems pointless, since if there are newlines in the Base64 form then line-by-line masking is not going to work anyway.)

src/test/java/org/jenkinsci/plugins/credentialsbinding/masking/Base64SecretPatternFactoryTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+public class Base64SecretPatternFactoryTest {
+
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    public static final String SAMPLE_PASSWORD = "}#T14'GAz&H!{$U_";
+
+    @Test
+    public void Base64SecretsAreMaskedInLogs() throws Exception {
```
jglick, line 20:
```suggestion
    public void base64SecretsAreMaskedInLogs() throws Exception {
```

src/main/java/org/jenkinsci/plugins/credentialsbinding/masking/Base64SecretPatternFactory.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private String removeTrailingEquals(String base64Value) {
+        if (base64Value.endsWith("==")) {
+            // removing the last 3 characters, the character before the == being incomplete
+            return base64Value.substring(0, base64Value.length() - 3);
+        }
+        if (base64Value.endsWith("=")) {
+            // removing the last 2 characters, the character before the = being incomplete
+            return base64Value.substring(0, base64Value.length() - 2);
+        }
```
jglick, line 57: Some tricky logic here. Can it be covered by a unit test? (Not `Base64SecretPatternFactoryTest` which is a functional test.)

src/main/java/org/jenkinsci/plugins/credentialsbinding/masking/Base64SecretPatternFactory.java
Kevin-CB: Good point, addressed by [d685136](https://github.com/jenkinsci/credentials-binding-plugin/pull/269/commits/d68513672d0287dca7bf97cbd62e1f89daa8eef5).

src/main/java/org/jenkinsci/plugins/credentialsbinding/masking/Base64SecretPatternFactory.java
Kevin-CB: I added some unit tests, do you think they're sufficient?

src/main/java/org/jenkinsci/plugins/credentialsbinding/masking/Base64SecretPatternFactory.java
```diff
@@ -0,0 +1,59 @@
...
+        if (secret.length() == 0) {
+            return Collections.emptyList();
+        }
+
+        Base64.Encoder[] encoders = new Base64.Encoder[]{
+                Base64.getEncoder(),
+                Base64.getUrlEncoder(),
+        };
+
+        Collection<String> result = new ArrayList<>();
```
yaroslavafenkin, line 33: Perhaps use Set over List?

yaroslavafenkin (commented):
Seems to work well in local testing

jglick (approved):

jglick:
Ready to go I think; @daniel-beck or @Wadeck or anyone else want to take a look?

Wadeck (approved):
Not manually tested this version, but looks very similar to what I expected 👍

src/main/java/org/jenkinsci/plugins/credentialsbinding/masking/Base64SecretPatternFactory.java
jglick: Should not matter I think; all the patterns will wind up merged into one big pattern sorted by length.

jglick:
Check https://github.com/jenkinsci/config-file-provider-plugin/pull/296.

PR #530, opened 2026-07-07
Skip credentials cleanup when the agent is offline

### LINKED ISSUE #531: withCredentials cleanup throws AgentOfflineException and fails the build when the agent is lost mid-block
### Jenkins and plugins versions report

<details>
<summary>Environment</summary>

- Jenkins: 2.555.2
- credentials-binding: 725.ve52b_2328a_fde
- workflow-durable-task-step: 1479.v56e587f413a_7
- workflow-cps: 4350.vcc65d4958821
- pipeline-model-definition: 2.2291.v2934911987b_6
- kubernetes: 4437.vd64141124d8f

</details>

### What Operating System are you using (both controller, and any agents)?

Linux controller; ephemeral Linux agents (Kubernetes pods). Reproducible with any agent that can be removed while a `withCredentials` block is active.

### Reproduction steps

1. Run a Pipeline on an agent with a `withCredentials` block that binds a file/secret, e.g.:
   ```groovy
   node('some-agent') {
     withCredentials([file(variable: 'SECRET', credentialsId: 'creds')]) {
       // long-running work
       sh 'sleep 600'
     }
   }
   ```
2. While the block is running, cause the agent to be lost without a graceful shutdown (e.g. delete/preempt the Kubernetes node so the pod is gone, or otherwise break the remoting channel so the node is later removed).

### Expected Results

The `withCredentials` cleanup (deleting the temporary secret files from the workspace) is best-effort. When the agent/workspace no longer exists, there is nothing to clean up, so the cleanup should be skipped without adding a failure.

### Actual Results

The deferred cleanup callback `BindingStep$Callback.finished` resolves the workspace via `context.get(FilePath.class)`, which throws `AgentOfflineException` because the agent is gone. The callback collects and re-throws it, so the build fails on the credentials cleanup:

```
org.jenkinsci.plugins.workflow.support.steps.AgentOfflineException: Unable to create live FilePath for <agent>; <agent> was marked offline: ...
	at ...ExecutorStepDynamicContext$FilePathTranslator.get(ExecutorStepDynamicContext.java:188)
	...
	at PluginClassLoader for credentials-binding//org.jenkinsci.plugins.credentialsbinding.impl.BindingStep$Callback.finished(BindingStep.java:270)
	at PluginClassLoader for credentials-binding//org.jenkinsci.plugins.credentialsbinding.impl.BindingStep$Execution2$Callback2.finished(BindingStep.java:184)
	at ...GeneralNonBlockingStepExecution$TailCall.lambda$onFailure$1(...)
```

This turns a *recoverable* agent loss into a hard error. Worse, under `retry` / declarative `agent { … retries N }`, this cleanup can fire late (after the "wait for the agent to come back" window) and surface as the **terminal build result**, so the retried attempt is defeated and the build fails referencing the already-dead agent.

### Anything else?

Root cause: `AgentOfflineException` (package-private, extends `IOException`) is raised while resolving the workspace `FilePath` in the cleanup callback. The unbind only performs `workspace.child(tempDir).deleteRecursive()` (`UnbindableDir.UnbinderImpl.unbind`), which is impossible and unnecessary once the workspace is gone.

Proposed fix: resolve `FilePath`/`Launcher` up front in `Callback.finished`; on `IOException` (agent offline/removed) log and return without throwing. Genuine unbind failures (agent reachable) still propagate.

### Are you interested in contributing a fix?

Yes — PR submitted.

wuhuizuo:
Fixes #531

### Problem

When an agent is lost while a `withCredentials` block is active (e.g. a Kubernetes pod is
deleted or a node is preempted), the deferred cleanup callback `BindingStep$Callback.finished`
resolves the agent workspace via `context.get(FilePath.class)`. Since the agent/channel is gone,
that throws `AgentOfflineException`, which the callback collects and re-throws.

Consequences:

- A *recoverable* agent loss is turned into a hard error.
- Under `retry` / declarative `agent { … retries N }`, this cleanup can fire late (after the
  "wait for the agent to come back" window) and surface as the **terminal build result**,
  defeating the retry — the build fails referencing the already-dead agent.

The cleanup only deletes the temporary secret files from the workspace
(`UnbindableDir.UnbinderImpl.unbind` → `deleteRecursive()`), which is impossible and pointless
once the workspace/agent is gone.

### Change

In `Callback.finished`, resolve `FilePath`/`Launcher` up front; if that fails with `IOException`
(the agent is offline/removed), log a line and return without throwing. Genuine unbind failures
(agent reachable) are still collected and thrown as before. Behavior for workspace-less bindings
is unchanged (`get(FilePath)` returns `null` rather than throwing).

Note: `AgentOfflineException` is package-private in `workflow-durable-task-step`, so it is caught
via its `IOException` supertype. If preferred, a follow-up could make that type public and catch
it specifically.

### Testing

- New `BindingStepTest#cleanupSkippedWhenAgentGoesOffline`: runs a `withCredentials` file binding
  on an agent, removes the node mid-body, and asserts the build logs
  `Skipping credentials cleanup because the agent is no longer online` instead of failing via the
  cleanup callback.
- Existing `BindingStepTest` suite still passes (19/19).
- Built and tested with Maven 3.9 / Java 21 against baseline Jenkins 2.479.3.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStep.java
@@ -34,6 +34,7 @@ import hudson.model.AbstractBuild;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.util.Secret;
+import java.io.IOException;
 import java.io.ObjectStreamException;
 import java.io.OutputStream;
 import java.io.Serializable;
@@ -263,11 +264,27 @@ public final class BindingStep extends Step {
         }
 
         @Override protected void finished(StepContext context) throws Exception {
+            final FilePath workspace;
+            final Launcher launcher;
+            try {
+                workspace = context.get(FilePath.class);
+                launcher = context.get(Launcher.class);
+            } catch (IOException x) {
+                // The agent that hosted the workspace is no longer online (e.g. AgentOfflineException:
+                // the build agent's pod/node was deleted or lost). The temporary credential files lived
+                // in that workspace and are gone with it, so there is nothing to unbind. Skipping cleanup
+                // here avoids turning a recoverable agent loss into a build failure (and, under retry,
+                // avoids this late cleanup becoming the terminal build result).
+                context.get(TaskListener.class).getLogger().println(
+                        "Skipping credentials cleanup because the agent is no longer online: " + x.getMessage());
+                return;
+            }
+
             Exception xx = null;
 
             for (MultiBinding.Unbinder unbinder : unbinders) {
                 try {
-                    unbinder.unbind(context.get(Run.class), context.get(FilePath.class), context.get(Launcher.class), context.get(TaskListener.class));
+                    unbinder.unbind(context.get(Run.class), workspace, launcher, context.get(TaskListener.class));
                 } catch (Exception x) {
                     if (xx == null) {
                         xx = x;
--- a/src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/BindingStepTest.java
@@ -312,6 +312,33 @@ class BindingStepTest {
         });
     }
 
+    @Test
+    void cleanupSkippedWhenAgentGoesOffline() throws Throwable {
+        final String secret = "s3cr3t";
+        extension.then(r -> {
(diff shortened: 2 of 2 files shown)


dillon-zheng (approved):
lgtm

jglick (changes requested):
> The cleanup only deletes the temporary secret files from the workspace […], which is impossible and pointless once the workspace/agent is gone.

On an _ephemeral_ agent such as a K8s pod, sure. Not so on a permanent agent.

In the associated issue I questioned the motivation for this.

jglick:
No response.

Issue #532, opened 2026-07-08
SSH and Certificate file can leave the expected directory


Kevin-CB:
### Jenkins and plugins versions report

Jenkins: 2.564-SNAPSHOT
OS: Windows 11 - 10.0
Java: 21.0.7 - Eclipse Adoptium (OpenJDK 64-Bit Server VM)

...
credentials-binding:725.ve52b_2328a_fde

### What Operating System are you using (both controller, and any agents involved in the problem)?

Windows, not bothering with agent, directly on controller.

### Reproduction steps

 `SSHUserPrivateKeyBinding` (`keyDir.getDirPath().child("ssh-key-" + keyFileVariable)`) and `CertificateMultiBinding` (`secrets.getDirPath().child("keystore-" + keystoreVariable)`) build the destination file path by concatenating a static prefix with the binding's variable name (`keyFileVariable` / `keystoreVariable`), which comes straight from the pipeline step's

If that variable name contains `../` (or `..\`) traversal sequences, the credential's secret file (SSH private key or certificate keystore) gets written outside the intended per-binding secret directory managed by `UnbindableDir`, instead of staying contained the way the other bindings ensure.

NOTE: This is **not a security issue**, since exploiting it requires the ability to author the Pipeline script itself (i.e. `Item/Configure`), a permission that already implies broad trust. 
Filing as a hardening fix to align these two bindings with the sanitization already applied to `FileBinding`/`ZipFileBinding`.

```groovy
node {
  withCredentials([sshUserPrivateKey(credentialsId:'demo-ssh-key',
    keyFileVariable:'../../../../myFile')]) { }
}
```

### Expected Results

Input is sanitized

### Actual Results

It's not sanitized and allows to leave the current directory

### Are you interested in contributing a fix?

Yes, I'll propose a patch

PR #533, opened 2026-07-08
Restrict file location for SSH and certificates credentials

Kevin-CB:
https://github.com/jenkinsci/credentials-binding-plugin/issues/532

### Testing done

Confirmed locally the file is not created outside the expected directory with the following pipeline
```
node {
  withCredentials([sshUserPrivateKey(credentialsId:'demo-ssh-key',
    keyFileVariable:'../../../../myFile')]) { }
}
```
did the same for certificate creds.

### Submitter checklist
- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [X] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding.java
@@ -1,5 +1,6 @@
 package org.jenkinsci.plugins.credentialsbinding.impl;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.security.KeyStoreException;
@@ -84,7 +85,8 @@ public class CertificateMultiBinding extends MultiBinding<StandardCertificateCre
 
 		if (workspace != null) {
 			final UnbindableDir secrets = UnbindableDir.create(workspace);
-			final FilePath secret = secrets.getDirPath().child("keystore-" + keystoreVariable);
+			String safeKeystoreVariable = new FilePath(new File(keystoreVariable)).getName();
+			final FilePath secret = secrets.getDirPath().child("keystore-" + safeKeystoreVariable);
 			OutputStream out = secret.write();
 			try {
 				credentials.getKeyStore().store(out, storePassword.toCharArray());
--- a/src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBindingTest.java
+++ b/src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBindingTest.java
@@ -30,6 +30,7 @@ import static org.hamcrest.CoreMatchers.nullValue;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 import java.io.File;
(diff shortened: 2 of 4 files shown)


src/test/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBindingTest.java
```diff
@@ -197,4 +198,29 @@ private static File newFile(File parent, String child) throws IOException {
...
+						"  }\n" +
+						"}", true));
+
+		r.buildAndAssertSuccess(p);
+
+		FilePath workspace = r.jenkins.getWorkspaceFor(p);
+		FilePath secretsDir = UnbindableDir.create(workspace).getDirPath();
+		FilePath resolvedPath = secretsDir.child("keystore-" + payload);
+
+		assertFalse(resolvedPath.exists(), "Keystore file should not exist outside secure directory");
```
daniel-beck, line 224: I find this lone assertion unconvincing. Didn't we do better in https://github.com/jenkinsci/credentials-binding-plugin/commit/3f6decef43ea5eebde2251ec5ea81b7eeeaa21ed already with the `#assertLogContains` positive assertion? Same in https://github.com/jenkinsci/credentials-binding-plugin/commit/e52b2328afdebcb69852687392fde94b174eaec4 followup. Why not add something similar here?

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding.java
```diff
@@ -84,7 +85,8 @@ public org.jenkinsci.plugins.credentialsbinding.MultiBinding.MultiEnvironment bi
 
 		if (workspace != null) {
 			final UnbindableDir secrets = UnbindableDir.create(workspace);
-			final FilePath secret = secrets.getDirPath().child("keystore-" + keystoreVariable);
+			String safeKeystoreVariable = new FilePath(new File(keystoreVariable)).getName();
```
daniel-beck, line 88: Looks familiar 👍  Also looks correctly adapted for the prefix unique to this case.

daniel-beck (commented):
Fix looks correct, but the tests could be improved.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding.java
```diff
@@ -84,7 +85,8 @@ public org.jenkinsci.plugins.credentialsbinding.MultiBinding.MultiEnvironment bi
 
 		if (workspace != null) {
 			final UnbindableDir secrets = UnbindableDir.create(workspace);
-			final FilePath secret = secrets.getDirPath().child("keystore-" + keystoreVariable);
+			String safeKeystoreVariable = new FilePath(new File(keystoreVariable)).getName();
+			final FilePath secret = secrets.getDirPath().child("keystore-" + safeKeystoreVariable);
```
jtnord, line 89: keystore variable is supposed to be the name of an environment variable not the temporary file!?

this should be a temporary file and not take in user specified paths.

e.g.

```
 withCredentials([certificate(credentialsId: 'id1', keystoreVariable: 'hello')]) {
 sh 'echo path is $hello and should exist'
   withCredentials([certificate(credentialsId: 'id2', keystoreVariable: 'hello')]) {
     sh 'echo path is $hello and should exist but be different to before'
   }
  sh 'echo path is $hello and should still exist (but the path for the previous id2 should be deleted'
}
```

So rather than attempting to sanitize somthing, this should surely just create a unique file without any user supplied being used.
(additionally - this does not sanitize against max filename lengths!)

jtnord (commented):
the variable should not even be used for file names, create a unique file in the directory.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding.java
jtnord, line 89:
```suggestion
			final FilePath secret = secrets.getDirPath().createTempFile("keystore-", ".ks"); 
```

jglick: I tend to agree. But: https://github.com/jenkinsci/credentials-binding-plugin/blob/f9d34fda028c3172c23d7199a87d797aab1ecc20/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/FileBinding.java#L54

Also https://github.com/jenkinsci/credentials-binding-plugin/blob/f9d34fda028c3172c23d7199a87d797aab1ecc20/src/main/java/org/jenkinsci/plugins/credentialsbinding/MultiBinding.java#L98-L99 should probably block anything that does not look like a valid variable name. (Too late for this attack.)

jglick: I think the original intent of using specific file names was to simplify diagnosis from build logs. But the bound file path is treated as a secret and masked anyway. So we might as well use arbitrary filenames (like random UUIDs).

jglick: `createTempFile` perhaps overkill given that each binding calls https://github.com/jenkinsci/credentials-binding-plugin/blob/f9d34fda028c3172c23d7199a87d797aab1ecc20/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/UnbindableDir.java#L52-L53 so even a fixed filename would work, but sure

jglick: ⇒ https://github.com/jenkinsci/credentials-binding-plugin/pull/534

Kevin-CB:
Closed in favor of https://github.com/jenkinsci/credentials-binding-plugin/pull/534

PR #534, opened 2026-07-08
Bind secret files to random temp paths

### LINKED ISSUE #532: SSH and Certificate file can leave the expected directory

jglick:
Fixes #532. Alternative to #533.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding.java
@@ -84,7 +84,7 @@ public class CertificateMultiBinding extends MultiBinding<StandardCertificateCre
 
 		if (workspace != null) {
 			final UnbindableDir secrets = UnbindableDir.create(workspace);
-			final FilePath secret = secrets.getDirPath().child("keystore-" + keystoreVariable);
+			final FilePath secret = secrets.getDirPath().createTempFile("file", null);
 			OutputStream out = secret.write();
 			try {
 				credentials.getKeyStore().store(out, storePassword.toCharArray());
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SSHUserPrivateKeyBinding.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SSHUserPrivateKeyBinding.java
@@ -102,7 +102,7 @@ public class SSHUserPrivateKeyBinding extends MultiBinding<SSHUserPrivateKey> {
                                            @NonNull TaskListener listener) throws IOException, InterruptedException {
         SSHUserPrivateKey sshKey = getCredentials(build);
         UnbindableDir keyDir = UnbindableDir.create(workspace);
-        FilePath keyFile =  keyDir.getDirPath().child("ssh-key-" + keyFileVariable);
+        FilePath keyFile =  keyDir.getDirPath().createTempFile("file", null);
 
         StringBuilder contents = new StringBuilder();
         for (String key : sshKey.getPrivateKeys()) {
--- a/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBinding.java
+++ b/src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBinding.java
@@ -32,7 +32,6 @@ import hudson.FilePath;
 import hudson.model.Item;
 import hudson.util.FormValidation;
 
-import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Collections;
@@ -58,9 +57,7 @@ public class ZipFileBinding extends AbstractOnDiskBinding<FileCredentials> {
     }
 
     @Override protected final FilePath write(FileCredentials credentials, FilePath dir) throws IOException, InterruptedException {
-        String baseName = new FilePath(new File(credentials.getFileName())).getName();
-
-        FilePath secret = dir.child(baseName);
+        FilePath secret = dir.createTempDir("file", null);
         secret.unzipFrom(credentials.getContent());
         secret.chmod(0700); // note: it's a directory
         return secret;
--- a/src/test/resources/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBindingTest/basicsPipeline-step2.sh
+++ b/src/test/resources/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBindingTest/basicsPipeline-step2.sh
@@ -5,7 +5,7 @@ set -e -x
 keystore_path=$(cat keystore-path)
 
 # check it was where we would expect it to be
-echo "$keystore_path" | grep -q '/workspace/p@tmp/secretFiles/[-0-9a-f]\{36\}/keystore-MY_KEYSTORE$'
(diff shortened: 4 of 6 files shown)


Kevin-CB (approved):
Yeah, using random temp is a better approach, there is no reason to involve user input in the ondisk filename at all.. it closes the vector more cleanly that sanitizing. 

I confiormed locally that was working properly. 

Closing https://github.com/jenkinsci/credentials-binding-plugin/pull/533 in favor of this one. 

Thanks Jesse!

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding.java
```diff
@@ -84,7 +84,7 @@ public org.jenkinsci.plugins.credentialsbinding.MultiBinding.MultiEnvironment bi
 
 		if (workspace != null) {
 			final UnbindableDir secrets = UnbindableDir.create(workspace);
-			final FilePath secret = secrets.getDirPath().child("keystore-" + keystoreVariable);
+			final FilePath secret = secrets.getDirPath().createTempFile("file", null);
```
jtnord, line 87: to aid debugging, use the type there (its a Certificate so will be a keystore, but could also be called `cert`) 

```suggestion
			final FilePath secret = secrets.getDirPath().createTempFile("keystore", null);
```

jtnord, line 87: alternative to above

```suggestion
			final FilePath secret = secrets.getDirPath().createTempFile("cert", null);
```

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SSHUserPrivateKeyBinding.java
```diff
@@ -102,7 +102,7 @@ public String getPassphraseVariable() {
                                            @NonNull TaskListener listener) throws IOException, InterruptedException {
         SSHUserPrivateKey sshKey = getCredentials(build);
         UnbindableDir keyDir = UnbindableDir.create(workspace);
-        FilePath keyFile =  keyDir.getDirPath().child("ssh-key-" + keyFileVariable);
+        FilePath keyFile =  keyDir.getDirPath().createTempFile("file", null);
```
jtnord, line 105:
```suggestion
        FilePath keyFile =  keyDir.getDirPath().createTempFile("id", null);
```

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBinding.java
```diff
@@ -58,9 +57,7 @@ public class ZipFileBinding extends AbstractOnDiskBinding<FileCredentials> {
     }
 
     @Override protected final FilePath write(FileCredentials credentials, FilePath dir) throws IOException, InterruptedException {
-        String baseName = new FilePath(new File(credentials.getFileName())).getName();
-
-        FilePath secret = dir.child(baseName);
+        FilePath secret = dir.createTempDir("file", null);
```
jtnord, line 60: use an extension for zips?

```suggestion
        FilePath secret = dir.createTempDir("file", "zip");
```

src/test/resources/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBindingTest/basicsPipeline-step2.sh
```diff
@@ -5,7 +5,7 @@ set -e -x
 keystore_path=$(cat keystore-path)
 
 # check it was where we would expect it to be
-echo "$keystore_path" | grep -q '/workspace/p@tmp/secretFiles/[-0-9a-f]\{36\}/keystore-MY_KEYSTORE$'
+echo "$keystore_path" | grep -q '/workspace/p@tmp/secretFiles/[-0-9a-f]\{36\}/file.\+[.]tmp$'
```
jtnord, line 8: depending on accepting previous suggestions:

```suggestion
echo "$keystore_path" | grep -q '/workspace/p@tmp/secretFiles/[-0-9a-f]\{36\}/keystore.\+[.]tmp$'
```

jtnord, line 8: or depending on accepting previous suggestions:
```suggestion
echo "$keystore_path" | grep -q '/workspace/p@tmp/secretFiles/[-0-9a-f]\{36\}/cert.\+[.]tmp$'
```

jtnord (approved):

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/ZipFileBinding.java
jglick: No because this is a _dir_ not a ZIP file.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBinding.java
jglick: I thought about it but settled on just a generic filename in all cases. As noted in #533, the actual filename is concealed from the logs anyway, so there is no real advantage to picking a particular prefix.

src/main/java/org/jenkinsci/plugins/credentialsbinding/impl/SSHUserPrivateKeyBinding.java
jglick: ditto

timja:
User report of issue in https://github.com/jenkinsci/credentials-binding-plugin/issues/539

rmontag-ap:
https://github.com/jenkinsci/credentials-binding-plugin/issues/536

PR #546, opened 2026-09-17
Re-resolve selected bound credentials per step (for rotating tokens, e.g. GitHub App)

### LINKED ISSUE #545: Option to re-resolve (not cache) bound credentials per step, for long builds where the credential rotates (e.g. GitHub App tokens)
## Problem

`withCredentials` / `environment { X = credentials(...) }` resolve each binding eagerly in `BindingStep` and store the plaintext in a static `Overrider` (`EnvironmentExpander`) that lives for the whole block; the masking `ConsoleLogFilter` is compiled once as well. For most credentials this is fine.

However, a **GitHub App** credential (from `github-branch-source`) issues an **installation token valid for 1 hour**, and that token is refreshed only when the credential is *resolved* (`GitHubAppCredentials` marks it stale ~45 min before expiry). In builds that run longer than the token TTL, the value captured at bind time goes **stale between steps**, and later Git/GitHub operations fail with 401/403 — even though re-resolving the credential would have produced a fresh token.

This is awkward to work around today because it requires editing every affected Jenkinsfile to narrow the `withCredentials` scope around each late Git/GitHub operation.

## Proposal A (opt-in, minimal surface)

A global configuration listing **credential IDs** whose bindings should **re-resolve on each `expand()`** instead of being cached, with **dynamic "union" masking** so rotated values stay masked on **both** the console and the exception/failure surface (the `MaskedException` that populates `ErrorAction` / `/wfapi` / `catch (e) { e.getMessage() }`, which does not go through the `ConsoleLogFilter`). An empty list means byte-for-byte the current behavior, so it is fully backwards compatible and off by default.

## Proposal B (cleaner, larger)

Introduce a small SPI so a credential *type* can declare itself "refreshable", letting `GitHubAppCredentials` opt in without an explicit ID allowlist.

## What we have

We have a working implementation of Proposal A as a fork of this plugin, currently running in production, with tests covering: per-step re-resolution, union masking, failure-surface masking, and CPS/agent serialization (durability across a controller restart). The stock code path is unchanged when the allowlist is empty (the upstream test suite stays green).

## Ask

Before opening a PR: which direction (A, B, or something else) would you be willing to accept for a security-sensitive core plugin? We are happy to submit the PR against `master` in whatever shape you prefer, with tests, and sign the CLA.

jmvega-pki:
## Summary

Adds an opt-in way to have selected credentials **re-resolved on every step** instead of cached for the whole lifetime of the `withCredentials` / `environment {}` block.

Motivated by GitHub App installation tokens (1h TTL, refreshed only when the credential is *resolved*): in builds longer than the token lifetime, the value captured at bind time goes stale **between steps**, and later Git/GitHub operations fail with 401/403. Today the only workaround is editing every affected Jenkinsfile to narrow the binding scope.

Implements "Proposal A" discussed in #545.

## What it does

- New global configuration ("Refresh bound credentials") with a **credential selector** — a repeatable dropdown of existing credentials (no free-text IDs).
- For the selected credentials, `BindingStep` installs a refreshing `EnvironmentExpander` that re-runs `MultiBinding.bind(...)` on each `expand()`, so each step gets a freshly resolved value (a fresh token when the credential has rotated).
- Dynamic "union" masking so rotated values stay masked on **both** the console **and** the failure/exception surface (the `MaskedException` that populates `ErrorAction` / `/wfapi` / `catch (e) { e.getMessage() }`, which does not pass through the `ConsoleLogFilter`).
- **Empty selection ⇒ byte-for-byte current behavior** — fully backwards compatible and off by default. Non-selected credentials keep the stock static `Overrider`/`Filter` path unchanged.

## Tests

Added coverage: gate (refreshing vs stock caching), per-step re-resolution, union masking, failure-surface masking, and CPS/agent serialization (durability across a controller restart). `mvn verify` is green and SpotBugs is clean; the existing suite is unchanged when nothing is selected.

## Notes

Happy to rework toward a "refreshable credential type" SPI (Proposal B from #545) if maintainers prefer that direction. I'll sign the CLA.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jglick (changes requested):
Complex change to a security-sensitive plugin; and not really the direction we would want anyway: https://github.com/jenkinsci/credentials-binding-plugin/issues/545#issuecomment-5733290738

jmvega-pki:
Thanks @jglick for the quick and clear feedback — much appreciated. Understood on both points (complexity in a security-sensitive plugin, and the preferred direction). We'll keep this as an internal fork rather than pursue it upstream, and we'll evaluate your suggestion of minting the installation token on demand from the App private key within the pipeline. Closing this PR.

